PR #4, opened 2014-06-06
Implement

matthiasbeyer:
This PR implements most hashtable functions.

Please review.


include/libreset/ht.h
```diff
@@ -38,15 +38,15 @@ void rs_ht_set_hasherfunc   (struct rs_ht *ht, rs_ht_hashfunc f);
...
+rs_ht_hash          rs_ht_put    (struct rs_ht *ht, void *d, size_t ds);
+void *              rs_ht_get    (struct rs_ht *ht, rs_ht_hash);
+void *              rs_ht_del    (struct rs_ht *ht, rs_ht_hash);
 
 /*
  * metainformations
  */
 
-rs_ht_bucket_id     rs_ht_size       (struct rs_ht *ht);
+rs_ht_hash          rs_ht_size       (struct rs_ht *ht);
```
matthiasbeyer: The `rs_ht_hash` type represents a size in here. I think this is somewhat inconsistent, but it is the simplest approach for it...

src/ht.c
```diff
@@ -72,16 +131,21 @@ rs_ht_del(struct rs_ht *ht, rs_ht_bucket_id)
...
-    rs_ht_bucket_id i;
+    rs_ht_hash s = 0;
+    rs_ht_hash i;
 
     for(i = 0; i < ht->length; i++) {
-        if (ht->buckets[i]) {
+        struct ll_element el = ht->buckets[i];
+
+        for(; el; el = el->next) {
+            /* if (el->value) { */
```
matthiasbeyer: This if is unneccessary as we do not use empty linkedlist-elements by now. But we need it if we do value-only removals inside the linkedlists.

Maybe we should introduce options for this.

PR #8, opened 2014-06-30
Add paper data layout

neithernut:
This patchset adds contents to the section "Data layout" which describes the internal structure of the set in more detail.

### DIFF
--- a/paper/data-layout.tex
+++ b/paper/data-layout.tex
@@ -77,3 +77,11 @@
         This feature will proof especially useful when performing resizes and
         binary operations on sets with different bucket-counts.
 
+    \subsection{Bucket sub-structure}
+
+        Each bucket consists of an AVL tree on it's own, using the hashes as
+        keys.
+        The AVL tree consists of nodes, each containing one or more elements
+        with identical hashes, e.g. a list of elements.
+
+
--- a/paper/data-layout.tex
+++ b/paper/data-layout.tex
@@ -17,3 +17,25 @@
     potentially be part of the bucket, rather than elements which \em are\em{}
     elements of the bucket.
 
+    \subsection{Number of buckets}
+
+        So far, we're just talking about an average hashtable. However, we
+        constrain the number of buckets $s$ to powers of two:
+
+        \begin{equation}
+            s \in \left\{ 2^n | n \in \mathbb{N} \right\}
+        \end{equation}
+
+        Assume having two hashtables $A$ and $B$, where the number of buckets of
+        table $B$ is greater or equal than that of table $A$.
+        The constraint ensures that, in this scenario, any bucket in $A$ is
+        equivalent to a subset of the buckets of $B$, regarding the elements it
+        may contain:
+
+        \begin{equation}
+            s(A) < s(B)
+            \quad\Rightarrow\quad
+            \forall_{0 \leq i \leq s(A)} : e(a_i) = \bigcup_{j\%s(B)=i} e(b_j)
+        \end{equation}
+
+
--- a/paper/data-layout.tex
+++ b/paper/data-layout.tex
@@ -1,4 +1,5 @@
 \section{Data layout}
+\label{sec:data_layout}
 
     As already implicated, the data is divided into several buckets --like in a
     regular hashtable. 
@@ -18,6 +19,7 @@
     elements of the bucket.
 
     \subsection{Number of buckets}
+    \label{sec:data_layout-bucketnum}
 
         So far, we're just talking about an average hashtable. However, we
         constrain the number of buckets $s$ to powers of two:
@@ -40,6 +42,7 @@
 
 
     \subsection{Division into buckets}
+    \label{sec:data_layout-bucketdiv}
 
         Naturally, a hashtable $H$ consists of a number of buckets $b_n$:
 
@@ -78,6 +81,7 @@
         binary operations on sets with different bucket-counts.
 
     \subsection{Bucket sub-structure}
+    \label{sec:data_layout-bucketstruct}
 
         Each bucket consists of an AVL tree on it's own, using the hashes as
         keys.
--- a/paper/data-layout.tex
+++ b/paper/data-layout.tex
@@ -39,3 +39,41 @@
         \end{equation}
 
 
+    \subsection{Division into buckets}
+
+        Naturally, a hashtable $H$ consists of a number of buckets $b_n$:
+
+        \begin{equation}
+            H = \left\{ b_n | n \in [0, s(H)) \right\}
+        \end{equation}
+
+        In regular hashtables, the bucket an element $i$ is put into is derived
+        by applying the modulo on the hash-value $h$ calculated from the
+        element:
+
+        \begin{equation}
+            n = h(i) \% s(H) \qquad i \in e(b_n)
+        \end{equation}
+
+        Taking into account the constraint described before, this would be
+        equivalent with taking the least significant bits of the hash-value.
+        However, we are using the \em most\em{} significant bits:
+
+        \begin{equation}
+            i \in e(b_n) \qquad n = h(i) \div \frac{2^N}{s}
+        \end{equation}
+
+        Where $N$ is the number of bits of a hash-value.
+        
+        Obviously, this preserves a subset of the natural numbers' ordering
+        relation: using this mapping, hashes of items in later buckets are
+        greater than those in earlier ones:
+
+        \begin{equation}
+            n_1 > n_2 \land i_1 \in e(b_{n_1}) \land i_2 \in e(b_{n_2})
+            \Rightarrow i_1 > i_2
+        \end{equation}
+
+        This feature will proof especially useful when performing resizes and
+        binary operations on sets with different bucket-counts.
+
--- a/paper/data-layout.tex
+++ b/paper/data-layout.tex
@@ -1,2 +1,19 @@
 \section{Data layout}
 
+    As already implicated, the data is divided into several buckets --like in a
+    regular hashtable. 
+
+    As to be expected, any element which may be part of the set is mapped onto
+    a bucket injectively.
+    The sets made up from the elements which would be mapped to a specific
+    bucket $e(b)$ are disjunctive to the elements which would be mapped to
+    another bucket of the same table:
+
+    \begin{equation}
+        e(b_1) \cap e(b_2) = \emptyset
+    \end{equation}
+
+    Note that $e(b)$ is the set of \em all\em{} elements which \em may\em{}
+    potentially be part of the bucket, rather than elements which \em are\em{}
+    elements of the bucket.
+
--- a/paper/main.tex
+++ b/paper/main.tex
@@ -6,6 +6,8 @@
 \usepackage[english]{babel}
 \usepackage{hyperref}
 
+\usepackage{amsfonts}
+
 
 \title{Fast set implementation based on a hybrid hashtable and AVL}
 \author{


paper/data-layout.tex
```diff
@@ -1,2 +1,79 @@
 \section{Data layout}
 
+    As already implicated, the data is divided into several buckets --like in a
```
matthiasbeyer, line 4: The `--like` results in bad formatted text.

paper/data-layout.tex
```diff
@@ -1,2 +1,79 @@
...
+        \end{equation}
+
+        Assume having two hashtables $A$ and $B$, where the number of buckets of
+        table $B$ is greater or equal than that of table $A$. The constraint
+        ensures that, in this scenario, the set of buckets in set $B$ which are
+        equivalent to a specific bucket in type $A$ is disjunctive to the set of
+        buckets in $B$ equivalent to another bucket in $A$:
+
+        \begin{equation}
+            n_{Buckets}(A) < n_{Buckets}(B) \land a_1 \neq a_2
```
matthiasbeyer: Isn't it `\rightarrow` instead of `\land`, as if A and B are not equal in size,
they are not equal?

paper/data-layout.tex
```diff
@@ -1,2 +1,79 @@
...
+            \Rightarrow
+            e(b) \subseteq e(a_1) \equiv \lnot e(b) \subseteq e(a_2)
+        \end{equation}
+
+        Where $a_t$ and $a_2$ buckets of $A$ and $b$ a bucket in $B$.
```
matthiasbeyer: There is a word missing in this sentence.

paper/data-layout.tex
```diff
@@ -1,2 +1,79 @@
...
+
+    \subsection{Division into buckets}
+
+        Naturally, a hashtable $H$ consists of a number of buckets $b_n$:
```
matthiasbeyer, line 47: We're getting a naming conflict here. The HT consists of buckets, but when
talking about the 'filled' buckets, we shouldn't call it `bucket`, too.

paper/data-layout.tex
```diff
@@ -1,2 +1,79 @@
...
+
+        \begin{equation}
+            H = \left\{ b_n | n \in [0, n_{Buckets}(H)) \right\}
```
matthiasbeyer: I assume it must be `n_{Buckets}(h)-1` in here, as if we have _one_ bucket, we would say here that n is in [0, 1], which would be two buckets. ...

paper/data-layout.tex
```diff
@@ -1,2 +1,79 @@
...
+        \end{equation}
+
+        In regular hashtables, the bucket an element $i$ is put is derived by
```
matthiasbeyer: Something is wrong here (wording).

paper/data-layout.tex
```diff
@@ -1,2 +1,79 @@
...
+        applying the modulo on the hash-value $h$ calculated from the element:
+
+        \begin{equation}
+            i \in e(b_n) \qquad n = h(i) \% n_{Buckets}(H)
```
matthiasbeyer: Would also be `n_{Buckets}(H)-1` here.

paper/data-layout.tex
```diff
@@ -1,2 +1,79 @@
...
+
+        However, we are using the \em most\em{} significant bits:
+
+        \begin{equation}
+            i \in e(b_n) \qquad n = h(i) \div \frac{2^N}{n_{Buckets}}
+        \end{equation}
+
+        Where $N$ is the number of bits of a hash-value.
+        
+        Obviously, this preserves a subset of the natural numbers' ordering
```
matthiasbeyer, line 71: I don't think that's obvious. This is my only comment related to the content: Please describe better here. I do not understand a single word! :-)

matthiasbeyer:
I commented on the diff, please fix the stuff in single commits and push them. I will ask you to rebase & squash _after_ I reviewed the next commits (this saves me from re-reviewing everything)!

matthiasbeyer:
Are you still alive, @neithernut ?

paper/data-layout.tex
neithernut: > Isn't it `\rightarrow` instead of `\land`, as if A and B are not
> equal in size, they are not equal?
> Nope, those two buckets not being equal is a second prequisite of the
> second statement after the \Rightarrow

paper/data-layout.tex
neithernut, line 47: > We're getting a naming conflict here. The HT consists of buckets, but
> when talking about the 'filled' buckets, we shouldn't call it
> `bucket`, too.
> I feel I have to come up with some naming convention and a more
> intuitive set of names for those predicates.

I'm on it.

paper/data-layout.tex
neithernut: > Would also be `n_{Buckets}(H)-1` here.
> I don't see why:
> 0%3=0
> 1%3=1
> 2%3=2

3%3=0
...

paper/data-layout.tex
neithernut, line 71: > I don't think that's obvious. This is my only comment related to the
> content: Please describe better here. I do not understand a single
> word! :-)
> Yes, I should make a few things more obvious.

paper/data-layout.tex
neithernut, line 4: > The `--like` results in bad formatted text.
> Could not verify.

paper/data-layout.tex
neithernut: > I assume it must be `n_{Buckets}(h)-1` in here, as if we have _one_
> bucket, we would say here that n is in [0, 1], which would be two
> buckets. ... 
> Nope, the range is exclusive of n_{Buckets} (it's a ")", not a "]").

paper/data-layout.tex
neithernut: > There is a word missing in this sentence.
> ...and it should be "a_1" rather than "a_t". Both issues are fixed now.

neithernut:
> I commented on the diff, please fix the stuff in single commits and
> push them. I will ask you to rebase & squash _after_ I reviewed the
> next commits (this saves me from re-reviewing everything)!
> As already noted somewhere else, there is still more work going into
> this, but hopefully I could clear some of the issues.

paper/data-layout.tex
neithernut: > `the bucket _is_ an`, missing word here?
> Nope, "the bucket an element $i$ is put" is the bucket, of which $i$
> will be an element.

paper/data-layout.tex
neithernut, line 71: But I seriously don't know how to make this more obvious. Maybe we should fix that one later.

neithernut:
I renamed a predicate, should make things more readable. Please review.

neithernut:
Btw: after the content is reviewed, I'd like to make each sentence start on it's own line. This way changes to a single sentence results in nicer commits.

neithernut:
This last commit concludes all there is to say about the data layout in this abstract form.

paper/data-layout.tex
matthiasbeyer, line 4: For me, compiling with `pdflatex main.tex`, the upper line results in

`[...] several buckets --like in [...]`

instead of

`[...] several buckets -- like in [...]`

Which I guess should be there. Notice the space after the double-dash (`--`)!

paper/data-layout.tex
matthiasbeyer, line 71: Ok. I will merge with this unfixed. (but the other stuff needs fixes anyways)

matthiasbeyer:
I have no time to review the overall content, _yet_. Will do this this evening.
The idea of having each sentence on one line: ACK, but I would stick to "each sentence one paragraph", like so:

```
foo bar
baz.
barbarz
bla.
```

(paragraphs without blank lines in between). So we don't use more than 80 characters per line!

neithernut:
Agreed. (was meant this way, anyway)

paper/data-layout.tex
```diff
@@ -1,2 +1,87 @@
...
+        equivalent to a specific bucket in type $A$ is disjunctive to the set of
+        buckets in $B$ equivalent to another bucket in $A$:
+
+        \begin{equation}
+            s(A) < s(B) \land a_1 \neq a_2
+            \Rightarrow
+            e(b) \subseteq e(a_1) \equiv \lnot e(b) \subseteq e(a_2)
+        \end{equation}
+
+        Where $a_1$ and $a_2$ are buckets of $A$ and $b$ a bucket in $B$.
```
matthiasbeyer: This sentence is syntactically wrong. It could also be embedded in the formula, which would be much more readable, as if you read the formula, you want to see something like this definition.

I would change the formula to:

``` latex
s(A) < s(B) \Rightarrow e(b) \subseteq e(a_1) \ equiv \lnot e(b) \subseteq e(a_2), a_n \elementof A, b \elementof B
```

or something similar.

---

I would also change the last part of the function (`e(b) \subseteq e(a_1) \equiv \lnot e(b) \subseteq e(a_2)`) to something like

``` latex
e(b) \subseteq e(a_1) \or \lnot e(b) \subseteq e(a_2)
```

(using an XOR)
As `e(b)` is _either_ a subset or equal to `a_1` _xor_ `a_2`, am I right?

I would contribute patches if you want me to change this,

paper/data-layout.tex
```diff
@@ -1,2 +1,87 @@
...
+
+    \subsection{Division into buckets}
+
+        Naturally, a hashtable $H$ consists of a number of buckets $b_n$:
+
+        \begin{equation}
+            H = \left\{ b_n | n \in [0, s(H)) \right\}
+        \end{equation}
+
+        In regular hashtables, the bucket an element $i$ is put is derived by
```
matthiasbeyer: `the bucket an element i is put ...`? There's one word missing (`into`) for me:

``` latex
the bucket an emlement $i$ is put into
```

paper/data-layout.tex
```diff
@@ -1,2 +1,87 @@
...
+        applying the modulo on the hash-value $h$ calculated from the element:
+
+        \begin{equation}
+            i \in e(b_n) \qquad n = h(i) \% s(H)
```
matthiasbeyer: Would change this to

``` latex
n = h(i) \% s(H), i \in e(b_n)
```

paper/data-layout.tex
```diff
@@ -1,2 +1,87 @@
...
+        However, we are using the \em most\em{} significant bits:
+
+        \begin{equation}
+            i \in e(b_n) \qquad n = h(i) \div \frac{2^N}{s}
+        \end{equation}
+
+        Where $N$ is the number of bits of a hash-value.
+        
+        Obviously, this preserves a subset of the natural numbers' ordering
+        relation: using this mapping, hashes of items in later buckets are
```
matthiasbeyer, line 72: What are "later buckets" ? (and "earlier buckets, of course)

paper/data-layout.tex
```diff
@@ -1,2 +1,87 @@
...
+
+        This feature will proof especially useful when performing resizes and
+        binary operations on sets with different bucket-counts.
+
+    \subsection{Bucket sub-structure}
+
+        Each bucket consists of an AVL tree on it's own, using the hashes as
+        keys.
+        The AVL tree consists of nodes, each containing one or more elements
+        with identical hashes, e.g. a list of elements.
```
matthiasbeyer, line 89: Do these elements _really_ have equivalent _hashes_? I don't think so! The hash can differ, but as we do an modulo-operation on the hash, the element results in the same bucket as another element, which can, of course, have another hash as the prior element!

I think this has to be changed in the text, as it leads to false assumptions.

paper/data-layout.tex
neithernut: There's actually a short form of the formular statement.

On the sentence: I would not embed it into the formula, as this would bloat things up.

neithernut: In fact, I'll rewrite the entire paragraph. It will express a different statement, which will imply the original one.

paper/data-layout.tex
neithernut: Added the "into"

paper/data-layout.tex
neithernut: Agreed.

paper/data-layout.tex
neithernut, line 72: Picturing an array of buckets, how is this not understandable. Or how could an order/positioning expressed more intuitively?

paper/data-layout.tex
neithernut, line 89: Hash-collisions do happen. (two items having, indeed, identical hashes)

paper/data-layout.tex
neithernut, line 4: I disagree.

neithernut:
Pushed a few updates.
Note: once all the blockers are resolved, I'd like to add labels, reformat the source and rebase.

matthiasbeyer:
The updates look good.

ACK for reformat, ACK for rebase (only rebasing to the new master, not for squash/fixup commits, as this step is trivial most of the time and this should be done _directly_ before merging).

neithernut:
Rebased.

paper/data-layout.tex
matthiasbeyer, line 72: ACK.

paper/data-layout.tex
matthiasbeyer, line 89: ACK. Understand something wrong here, got it after re-reading.

neithernut:
Source reformat and minor fix. Do I have a "go" for rebasing/squashing?

matthiasbeyer:
Yes, you have.

neithernut:
Done.

paper/data-layout.tex
```diff
@@ -1,2 +1,91 @@
 \section{Data layout}
+\label{sec:data_layout}
+
+    As already implicated, the data is divided into several buckets --like in a
```
matthiasbeyer, line 4: This still results in bad formatted text ("several buckets -like in a" instead of "several buckets - like in a").

neithernut, line 4: This was done on purpose and I do not consider this a formatting error. See http://en.wikipedia.org/wiki/Dash

matthiasbeyer, line 4: Then, this set is fine.

Merged.

PR #9, opened 2014-07-02
Fix typos

matthiasbeyer:
This PR fixes some typos. Please review & ACK.

### DIFF
--- a/paper/overview.tex
+++ b/paper/overview.tex
@@ -12,6 +12,6 @@
     derivation of the operations class of complexity.
 
     Additionally, we will discuss the possible use of bloom filters within this
-    data-structure for early denyablity of an elements occurrence in the set.
+    data-structure for early deniablity of an elements occurrence in the set.
 
 
--- a/paper/overview.tex
+++ b/paper/overview.tex
@@ -8,8 +8,8 @@
 
     This paper will discuss the algorithms used to implement each operation,
     drawing lines between the different levels, the overall hash-table and the
-    buckets that is. Each operation's description will also include a short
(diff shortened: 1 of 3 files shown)


paper/overview.tex
```diff
@@ -8,10 +8,10 @@ \section{Overview}
 
     This paper will discuss the algorithms used to implement each operation,
     drawing lines between the different levels, the overall hash-table and the
-    buckets that is. Each operation's description will also include a short
-    derivation of the operation's class of complexity.
+    buckets that is. Each operations description will also include a short
+    derivation of the operations class of complexity.
```
neithernut, line 13: It's a description of the operation, hence the operation's description, same for the second line.

matthiasbeyer, line 13: But in english, you use `'s` as shortcut for `is`. So after expanding, the upper says:

`Each operation is description will also include...`

Which isn't right, is it?

neithernut:
It depends on the context. An apostrophe, followed by an "s" can be either an "it" or an indicator of possession or affliliation.

matthiasbeyer:
Please re-add this comment inside the diff discussion and delete the old one. I will delete this comment afterwards.

paper/overview.tex
TheNeikos, line 13: 's is still right here, it merely means that the description belongs to the operation. It doesn't always expand to `is`

matthiasbeyer, line 13: Ok. Will merge then.

matthiasbeyer:
Wha shit, I'm to tired this morning. I will submit a revert-patch!

PR #10, opened 2014-07-03
Added astylerc and pre-commit hook

matthiasbeyer:
This PR adds a `astylerc` file for formatting with astyle.

I'm not sure whether to add a pre-commit checker or something, too.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -44,9 +44,8 @@ do
     cp $file $fpath
 
     astyle --options=./etc/astylerc < $fpath > $fpath_formatted
-    diff $fpath $fpath_formatted > /dev/null
 
-    if [ $? -ne 0 ]
+    if ! diff $fpath $fpath_formatted > /dev/null
     then
         echo_red "Format bad: $file"
         bad=1
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -10,6 +10,9 @@ do
     fi
 done
 
+p=$(pwd)
+cd $(git rev-parse --show-toplevel)
+
 Green="\033[32m"
 Red="\033[31m"
 Cyan="\033[0;36m"
@@ -65,7 +68,9 @@ rm -rf $tmpdir
 if [ $bad -eq 1 ]
 then
     echo_red "Bad formatted file found. You cannot commit"
+    cd $p
     exit 1
 fi
 
+cd $p
 exit 0
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,16 +1,13 @@
 #!/bin/bash
 
-if [ -z $(which astyle 2>/dev/null) ]
-then
-    echo "No 'astyle' found, please install or put in PATH"
-    echo "You cannot commit"
-fi
-
-if [ -z $(which diff 2>/dev/null) ]
-then
-    echo "No 'diff' found, please install or put in PATH"
-    echo "You cannot commit"
-fi
+for i in astyle diff
+do
+    if ! type $i 2> /dev/null
+    then
+        echo "No '$i' found, please install or put it in PATH"
+        echo "You cannot commit"
+    fi
+done
 
 Green="\033[32m"
 Red="\033[31m"
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -44,6 +44,7 @@ do
     cp $file $fpath
 
     astyle --options=./etc/astylerc < $fpath > $fpath_formatted
+    # delete last line of $fpath_formatted
     sed -i '$ d' $fpath_formatted
 
     if ! diff $fpath $fpath_formatted > /dev/null
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -44,6 +44,7 @@ do
     cp $file $fpath
 
     astyle --options=./etc/astylerc < $fpath > $fpath_formatted
+    sed -i '$ d' $fpath_formatted
 
     if ! diff $fpath $fpath_formatted > /dev/null
     then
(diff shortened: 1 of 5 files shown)


neithernut:
> +--break-blocks

Do we want this?
What about, say:

| int i=8;
| while (i) {
|     //...
|     i>>=1;
| }

matthiasbeyer:
This is a good point. Shall I remove the `--break-blocks` line?

neithernut:
Is there an option which enforces spaces _after_ blocks only?

matthiasbeyer:
Not for astyle, sadly.

neithernut:
Then it should be removed, for now.

matthiasbeyer:
ACK?

matthiasbeyer:
I added a pre-commit hook, which solves #21 . Please review, @neithernut , @TheNeikos !

matthiasbeyer:
If we want to merge this, including the pre-commit hook which was added in this PR, we should do the following:
- Run `astyle` on each file which is in this branch atm, add the changes to _this_ PR
- Rebase all branches onto master after _this_ branch is merged
- Redo all commits in the other PRs, so we only add formatted code

Thoughts?

I would do this ASAP, so we don't get into trouble later on.

neithernut:
ACK.
Astyle-checking might get integrated into the cmake-setup later on, but it's fine for now.

matthiasbeyer:
Have you tested it? I tested it but maybe another test run would be fine!

If you don't want to mess up, just create a new clone of some C-code repository and run the hook when test-committing something.

neithernut:
I'll do this after I've processes all the other notifications in my inbox.

neithernut:
Mh... on this machine, bash is in /bin, not in /usr/bin

matthiasbeyer:
Is there a unified path for bash somewhere? So we can get the shebang line "platform independent" somehow?

neithernut:
I would have to read up on the POSIX standart, but sh was supposed to go into /bin (or be symlink'ed), I think. The most sane thing would be to port the script to sh.

matthiasbeyer:
@manuelmessner Do you want to port the script to (pure) sh?

manuelmessner:
> Is there a unified path for bash somewhere? So we can get the shebang line "platform independent" somehow?

/bin is the standard path. On arch linux bash is in /usr/bin but /bin is linked to /usr/bin due to compatibilty reasons.

> @manuelmessner Do you want to port the script to (pure) sh?

Yes i can do it.

matthiasbeyer:
So then please do it. Contact me directly for workflow questions.

matthiasbeyer:
@neithernut please review & test again. Issues should be fixed now.

hooks/pre-commit
```diff
@@ -0,0 +1,70 @@
...
+
+for file in $(git diff --name-only HEAD | grep "\.[ch]$")
+do
+    echo_cyan "Checking: $file"
+    fpath=$tmpdir/$(basename $file)
+    fpath_formatted=$tmpdir/$(basename $file)-formatted
+
+    cp $file $fpath
+
+    astyle --options=./etc/astylerc < $fpath > $fpath_formatted
```
neithernut, line 49: Seems like this line produces a newline at the end of the output file.
Running diff afterwards yields a non-null result...

matthiasbeyer, line 49: @manuelmessner do you want to solve this issue?

matthiasbeyer:
@neithernut Test again, please. I tested and the latest patches solve the issue you mentioned (https://github.com/matthiasbeyer/libreset/pull/10#discussion_r16911344) for me.

hooks/pre-commit
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
```
neithernut, line 2: The script currently only works when committing from the repo's root directory.
To enhance this, I would inset a "cd" to the path returned by "git rev-parse --show-toplevel" here.

matthiasbeyer, line 2: I wouldn't be sure about this. Did you actually test it? I would bet that git hooks are always executed from the repositories root directory!

neithernut, line 2: I tested the script directly.
In the past, I _had_ problems with commit-hooks _not_ being executed from the repo's root. So, even if this behaviour is fixed in new versions of git, including it doesn't hurt.

hooks/pre-commit
```diff
@@ -0,0 +1,76 @@
...
+rm -rf $tmpdir
+
+if [ $bad -eq 1 ]
+then
+    echo_red "Bad formatted file found. You cannot commit"
+    cd $p
+    exit 1
+fi
+
+cd $p
```
neithernut, line 75: Should not be neccessary, but it doesn't hurt either.

neithernut:
Seems to work just fine now.

matthiasbeyer:
So I can merge this?

neithernut:
Yep, but I think you should squash a few things first.

TheNeikos:
Why not use the bash builtins `pushd` and `popd` ?

matthiasbeyer:
@neithernut Rebased. I moved the "run astyle on <file>" commits to the end of the branch. As we do not have more `[ch]` files then the two mentioned, this PR should be ready now.

edit: Just noticed: Github does not show the history as-is. The "run 'astyle'..." commits are acutally at the end of the history!

matthiasbeyer:
@TheNeikos Feel free to submit another PR (which can be based on master after the merge).

TheNeikos:
Alright

matthiasbeyer:
For documentation: Closes #21

neithernut:
ACK

manuelmessner:
> Why not use the bash builtins pushd and popd?

These bulitins are not available in pure shells. 
And this would be the first compatiblity problem in the script.

PR #11, opened 2014-07-23
Define interface

matthiasbeyer:
This adds the definition of the public interface. Comments welcome!

@neithernut


include/libreset/set.h
```diff
@@ -0,0 +1,181 @@
...
+
+
+/**
+ * Set configuration type
+ */
+struct r_set_cfg
+{
+    size_t          (*hashf)(void const*); //!< hash function
+    int             (*cmpf)(void const*, void const*); //!< compare function
+    void const*     (*copy)(void const*); //!< copy function
```
TheNeikos: All the other functions take an f in the name, I think this one should too, something like "cpyf"

:smile:

include/libreset/set.h
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Insert an object into a set
+ *
+ * @return zero on success, else error code
+ */
+int
+r_set_insert(
+        struct r_set* set, //!< the set
+        void const* value //!< value to insert
```
TheNeikos: You actually insert a pointer to a value, not the value itself

include/libreset/set.h
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Check if two sets are equal
+ *
+ * @note Checks metainformation of the set objects first (comparator function,
+ * hashing function). If they are not equal, the function returns false (zero).
+ *
+ * @warning returns error codes on failure, which are true, too!
+ *
+ * @return 1 if the sets are equal, else 0 (zero)
```
TheNeikos, line 150: This is contradictory to what is written in the warning, you should also specify that error codes are returned here

include/libreset/set.h
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Select entries from a set for a new set
+ *
+ * @return zero on success, else error code
+ */
+int
+r_set_select(
+        struct r_set* dest, //!< destination set
+        struct r_set const* src, //!< source set
+        int (*predicate)(void const*, void*), //!< predicate for selection
```
TheNeikos, line 183: What does the functions signature mean here? I think there is something missing in the documentation

include/libreset/set.h
matthiasbeyer: Yes, so where is the problem? The value is const here, if I'm right... we may discuss this anyway, as @neithernut has some concerns about the constness of the values we store.

include/libreset/set.h
matthiasbeyer, line 150: You mean I should mention the error codes in the `@return` statement instead of the `@warning` statement?

include/libreset/set.h
matthiasbeyer, line 183: We select entries from one set and put it into another set, where's the problem? The predicate is the  function which selects or rejects an entry!

TheNeikos, line 183: But what are the arguments for the predicate function supposed to mean?

The first is a pointer on a constant void, and the second just a void pointer. What is each one of them doing?

include/libreset/set.h
TheNeikos, line 150: You should mention them in both imo

include/libreset/set.h
TheNeikos: I meant the documentation, you wrote `value to insert` but it's actually just the pointer to a value.

Feeling a bit nitpicky :tongue:

include/libreset/set.h
matthiasbeyer, line 150: Yes, but this is a minor fix which can go in later.

include/libreset/set.h
matthiasbeyer, line 183: Of course, here is documentation missing. I'll add a fix.

include/libreset/set.h
matthiasbeyer: Well, you insert the pointer. We do not store the value, but a pointer to it! So... yes.

neithernut:
Can be merged imho.

PR #12, opened 2014-07-30
Add AVL implementation

matthiasbeyer:
Hi,

we need a AVL implementation, so I started defining an interface for one. Please review!

@neithernut , @theneikos , @manuelmessner

### DIFF
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -270,6 +270,37 @@ find_leftmost_node_with_parent(
 }
 
 /**
+ * Find a node by its hash
+ *
+ * @warning `root` may not be NULL
+ *
+ * @return Pointer to a node with the hash `hash` or NULL if none is found
+ *
+ * @todo Check relation
+ */
+static struct avl_el*
+find_node_with_hash(
+    struct avl_el* root,    //!< The node where to begin to search in
+    rs_hash h               //!< The hash to search for
+) {
+    if (root->hash == h) {
+        return root;
+    }
+
+    if (root->hash >= h) {
+        if (root->l) {
+            return find_node_with_hash(root->l, h);
+        }
+    } else {
+        if (root->r) {
+            return find_node_with_hash(root->r, h);
+        }
+    }
+
+    return NULL;
+}
+
+/**
  * Add a data element to a avl tree
  *
  * @return Ptr to the struct avl_el node where the data was inserted into or
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -81,6 +81,19 @@ find_higher_child(struct avl_el* root) {
 }
 
 /**
+ * Get the height difference between two nodes
+ */
+static inline signed int
+height_diff(
+    struct avl_el* a, //!< The first node
+    struct avl_el* b  //!< The first node
+) {
+    int h1 = (a == NULL ? 0 : a->height);
+    int h2 = (b == NULL ? 0 : b->height);
+    return ((h1 >= h2) ? (h1 - h2) : (h2 - h1));
+}
+
+/**
  * Perform a rotation between a root node and its left child
  *
  * @warning Call only if root node has left child
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -61,7 +61,24 @@ avl_del(
     int (*cmp)(void* const, void* const, void*),
     void* etc
 ) {
-    // TODO
+    struct avl_el* parent;
+    struct avl_el* found;
+    int f = find_element_with_parent(root, cmp, etc, &parent, &found);
+
+    if (f) {
+
+        // TODO: reorganize tree if neccessary, else:
+        if (found->l == NULL && found->r == NULL) {
+            if (parent->l == found) {
+                parent->l = NULL;
+            } else {
+                parent->r = NULL;
+            }
+        }
+
+        free(found);
+    }
+
     return 0;
 }
 
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -199,9 +199,7 @@ avl_add(
         new->l = NULL;
         new->r = NULL;
 
-        if (root) {
-            // TODO: Adding the element (new) to the tree (root)
-        }
+        new = insert(new, root);
     }
 
     return new;
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -27,32 +27,46 @@ find_element_with_parent(
     struct avl_el** parent_dest, //!< destination of the ptr to the parent
     struct avl_el** found_dest //!< destination of the ptr to the element
 ) {
-    struct avl_el* next;
-    int predret = pred(root, etc);
 
-    if (root == NULL) {
+    if (!root) {
         return 0;
     }
 
-    if (predret == 0) {
+    if (pred(root, etc) && found_dest) {
         if (parent_dest) {
             *parent_dest = NULL;
         }
-        if (found_dest) {
-            *found_dest = root;
-        }
+        *found_dest = root;
         return 1;
-    } else if (predret == -1) {
-        next = root->l;
-    } else {
-        next = root->r;
     }
 
-    if (parent_dest) {
-        *parent_dest = root;
+    if (find_element_with_parent(root->l, pred, etc, parent_dest, found_dest) ||
+        find_element_with_parent(root->r, pred, etc, parent_dest, found_dest)) {
+        /**
+         * Element FOUND in one of the subtrees tree
+         *
+         * - found_dest was set by find_element_with_parent() call
+         * - set parent_dest if and only if it was not set before (by a
+         *   recursion step)
+         */
+        if (parent_dest && *parent_dest != NULL) {
+            *parent_dest = root;
+        }
+
+        return 1;
     }
-    return find_element_with_parent(next, pred, etc, parent_dest, found_dest);
+    /* element NOT FOUND in one of the subtrees */
+
+    /**
+     * ensure the parent_dest is NULL
+     */
+    if (parent_dest) {
+        *parent_dest = NULL;
+    }
+
+    return 0;
 }
+
 
 /**
  * Get the higher child of a node
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -94,8 +94,9 @@ avl_find(
     int (*cmp)(void* const, void* const, void*),
     void* etc
 ) {
-    // TODO
-    return NULL;
+    struct avl_el* found;
+    find_element_with_parent(root, cmp, etc, NULL, &found);
+    return found;
 }
 
 struct avl_el*
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -311,13 +311,13 @@ avl_add(
  */
 struct avl_el*
 avl_del(
-    struct avl_el* root,
+    struct avl* avl,
     int (*pred)(void* const, void*),
     void* etc
 ) {
     struct avl_el* parent;
     struct avl_el* found;
-    int f = find_element_with_parent(root, pred, etc, &parent, &found);
+    int f = find_element_with_parent(avl->root, pred, etc, &parent, &found);
 
     if (f) {
 
@@ -404,7 +404,7 @@ avl_del(
      * @todo rebalancing, with rotations if neccessary.
      */
 
-    return root;
+    return avl->root;
 }
 
 struct avl_el*
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -4,7 +4,7 @@
 
 struct avl_el*
 avl_alloc(void) {
-    return avl_add(NULL, NULL);
+    return avl_add(NULL, NULL, NULL);
 }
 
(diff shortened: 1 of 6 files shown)


src/libreset/avl.h
```diff
@@ -0,0 +1,78 @@
+#ifndef __AVL_H__
+#define __AVL_H__
+
+struct avl_el {
+    void*   d;
```
TheNeikos: If this is data, please rename it as such, or documentation

src/libreset/avl.h
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Add an entry in an avl tree
+ *
+ * @return The new added element or NULL on failure
+ */
+struct avl_el*
+avl_add(
+        struct avl_el* root, //!< The root of the avl tree where to insert
+        int (*cmp)(void* const, void* const), //!< the comperator function
+        void* const d //!< The data element
```
TheNeikos: I just notived that there is also this one here that would need changing if we want to stay consistent.

matthiasbeyer: I guess this diff is already outdated. I just fixed this.

src/libreset/avl.h
```diff
@@ -0,0 +1,84 @@
...
+);
+
+/**
+ * Delete an element from the avl tree
+ *
+ * @return 1 on success, else a negative error code
+ */
+int
+avl_del(
```
neithernut, line 102: We may want to return a pointer to the data here, for optional further usage.
If the element was not found, NULL may be returned.

neithernut:
The comparator's return value has to be defines. Do we return [-1,0,1](preferable) or [0,1]?
Hint: this is a perfect example of when to use a typedef for  a function type.

Note: we will use the hash as a key inside the AVL tree, so supplementing a function isn't neccessary anyway.

matthiasbeyer:
So you mean we do not use cmp-functions? So should I remove the `cmp` parameter from the functions
- `avl_add`
- `avl_del`
- `avl_find`
- `avl_unlink`

?

neithernut:
Yep

matthiasbeyer:
I'll do this now, but we need some helper functions anyways, for the `avl_del()`, we need something which determines which element gets removed.

matthiasbeyer:
Anyways, this needs further patching, as the actual algorithms are not implemented.

Also, this needs thread-safeness.

src/libreset/util/macros.h
```diff
@@ -0,0 +1,14 @@
...
+
+#define MAX(x,y)                    \
+       ({ __typeof__ (x) _x = (x);  \
+          __typeof__ (y) _y = (y);  \
+          _x > _y ? _x : _y; })
+
+#define MIN(x,y)                    \
+       ({ __typeof__ (x) _x = (x);  \
+          __typeof__ (y) _y = (y);  \
+          _x < _y ? _x : _y; })
```
matthiasbeyer: These should be typesafe because we actually use variables in here.

We use a block `({...})`, so the temporary variables get out of scope after the macros are used.

matthiasbeyer:
I guess I need help on this. I guess we should do this together.

Remaining:
- Implement deleting
- Make threadsafe (I have no experience on this at all)

Do we need algorithms for merging two AVLs? If yes, this would be another point.

src/libreset/avl.h
```diff
@@ -0,0 +1,137 @@
...
+        void* const d, //!< The data element
+);
+
+/**
+ * Delete an element from the avl tree
+ *
+ * The element of the delete operation is returned, so we do not have a dangling
+ * pointer laying around.
+ *
+ * @return the element on success, else NULL
```
matthiasbeyer, line 99: I think we should better return the new root node, instead of the element. This sounds much more obvious, as with the rotation, the root can change!

matthiasbeyer:
This set gets huge. I think we should refactor the whole set after the implementation is ready, to shorten the history a bit, if possible.

And if you want, of course, because I don't think it is that urgent.

---

Edit:

Maybe we should re-roll this and add a `devel` and a `devel-avl` branch, whereas the latter is for developing the AVL implementation?
This way, there would be the possibility to get the utils which are added by this set (macros.h, element.h, hash.h) into the `devel` branch beforehand.

@neithernut , @TheNeikos thoughts?

TheNeikos:
Yes to the second one.

I am not so sure about the benefits of splitting up the branches other than a different view.

matthiasbeyer:
Well, of course it is a benefit mainly for the view.

---

Besides: You are (all) welcome to submit PRs onto this branch! When doing a PR, you can select where this PR should be merged, and you can simply define this branch as merge-point. So we can collaborate on one branch.
I would then do a `--ff` merge (offline, github can detect this), as this branch will be rebased (due to #10 ) anyway!

matthiasbeyer:
Some commits can be removed after #24 , #25 and #26  are merged.

neithernut:
Regarding merging and splitting functions: for now, construction/initialisation, destruction insertion and deletion should be enough. All the other things may go into another PR.

matthiasbeyer:
:+1:

matthiasbeyer:
Updated to latest master, including removing commits which were added to master already by other PRs already (mainly `src/libreset/hash.h`).

And removed `src/libreset/element.h` from the history by reverting and deleting appropriate patches.

src/libreset/avl.c
```diff
@@ -0,0 +1,421 @@
...
+        struct avl* avl,
+        int (*pred)(void* const, void*),
+        void* etc
+) {
+    struct avl_el* parent;
+    struct avl_el* found;
+    int f = find_element_with_parent(avl->root, pred, etc, &parent, &found);
+
+    if (f) {
+
```
neithernut, line 496: In such cases, I prefer an early return. The program flow is much clrearer this way imho and you save some intendations.

src/libreset/avl.c
```diff
@@ -0,0 +1,421 @@
...
+        /*
+         * Nothing should now point to `found` and nothing should dangling
+         * around, all ptrs from `found` to somewhere are safed.
+         */
+        free(found);
+    }
+
+    /*
+     * @todo rebalancing, with rotations if neccessary.
+     */
```
neithernut, line 561: The rebalancing method will be called by the higher level methods of the set.

src/libreset/avl.c
```diff
@@ -0,0 +1,421 @@
...
+
+/**
+ * Add a node to a tree
+ *
+ * @return the new root node
+ */
+static struct avl_el*
+insert(struct avl_el* new, struct avl_el** root)
+{
+    signed int cmpres = new->hash == root->hash ? 0 : (new->hash > root->hash ? -1 : 1);
```
neithernut: Shouldn't it be (*root)->hash?
Does this even compile?
If the computation of cmpres invlolves the trinary operator, it's pointless to not put the contidion right into the expression at the "if" itself anyway.

src/libreset/avl.c
```diff
@@ -0,0 +1,421 @@
...
+
+    if (*root == NULL) {
+        return new;
+    } else if (cmpres == 1) {
```
neithernut: This chunk is malicious, or malformed, I think.
You should check for *root==NULL first (returning if NULL) and determine cmpres _afterwards_
Has the advantage of not needing the else

src/libreset/avl.c
matthiasbeyer, line 561: So there will be no rebalancing in this function?

src/libreset/avl.c
matthiasbeyer: Thank you for reviewing this carefully! The rebase messed this up. Will add a fixup patch, of course.

src/libreset/avl.c
neithernut, line 561: nope

matthiasbeyer:
Will rebase (and merge the fixup commits in) after #10 is merged.

matthiasbeyer:
I rewrote all commits to include a pre-commit hook run (run astyle after each commit). It was fiddeling around, but I guess it worked.

matthiasbeyer:
Rebased and rewrote history due to #31 .

src/libreset/avl.c
```diff
@@ -0,0 +1,486 @@
...
+) {
+    int h1 = (a == NULL ? 0 : a->height);
+    int h2 = (b == NULL ? 0 : b->height);
+    return ((h1 >= h2) ? (h1 - h2) : (h2 - h1));
+}
+
+/**
+ * Recompute the height for a subtree
+ *
+ * @invariant \f$ -1 \leqslant H(n) \leqslant +1 \f$, Where \f$ H(n) \f$ is the
```
matthiasbeyer: Note: Requires "amssymb" latex package to be included for the doxygen documentation

matthiasbeyer:
@neithernut please review and tell me what needs to be done in this PR. @manuelmessner and @TheNeikos you're welcome, too!

matthiasbeyer:
With #35 , things get more complicated with the implementation this PR ships. Things have to be reworked. I would like to see some PRs onto this branch transforming it to the use of the linkedlist implementation ( #35 ).

neithernut:
Wait... you mean reimplement the insertion function you already wrote?

matthiasbeyer:
Not entirely.

The insertion algorithm is implemented for an element for which the hash is already in the tree. It is unimplemented for a node where the hash is not in the tree already.

Edit: I just can't wrap my head around the whole rotation foo. I know how it works, but I don't know when to  rotate left and when right, leave alone single and double rotations!

---

Edit: I can try, of course...

neithernut:
If you have real problems, strip the implementation and push the interface together with stubs.
We may then open an issue for each function not implemented and assign functions to people.
Rotations have, of course, be implemented first.

matthiasbeyer:
Rotations are already implemented!

src/libreset/avl.h
neithernut, line 99: Yep.

matthiasbeyer:
Rebased and squashed fixup commits.

neithernut:
To give the travic CI integration test some meaning, the source files have to be included in the CMake setup.

matthiasbeyer:
You are the expert on this, please add an appropriate PR onto this branch about this.

Isn't this covered by the `add_subdirectory(src)` line in the `CMakeLists.txt` file?

neithernut:
Nope, there is a second CMakeLists.txt in src. If you open that one for editing, it should be obvious to you what you have to do.

matthiasbeyer:
@neithernut How to add an include path? The build fails because the appropriate header wasn't found!

(`/home/travis/build/matthiasbeyer/libreset/src/libreset/avl.c:1:26: fatal error: libreset/avl.h: No such file or directory`)

neithernut:
But you still have to add the source files to `SOURCE_FILES` (line 12 in src/CMakeLists.txt)

matthiasbeyer:
@neithernut It lists the `avl.c` file, doesn't it? :-P

matthiasbeyer:
For documentation: due to #46 , the code comment converting commits (cccc - lol!) can and will be removed.

neithernut:
Ahh, I looked at an old commit.

matthiasbeyer:
@neithernut : Would you have a look at [the travis build](https://travis-ci.org/matthiasbeyer/libreset/builds/34787455), it still complains about ISO-C.

neithernut:
The warning message is clear: ISO-C forbids blocks in expressions.
Either the MAX macro has to be simplyfied to something likee:

``` C
#define MAX(a,b) ((a)>(b)?(a):(b))
```

or we do _not_ write ISO-C.

matthiasbeyer:
The idea of the macro is that it does a type safe comparison. You cannot do 

``` C
int a = 1;
double b = 2.0;
int c = MAX(a, b);
```

without getting a compiler warning about it, if we use the macro as it is defined. The simplification wouldn't be typesafe...

matthiasbeyer:
I would ACK an appropriate patch to the build environment for fixing this rather as to a patch changing the macros so they do not contain blocks.

Macros with type safety are much more of an improvement than ISO-C vs. non-ISO-C.

---

(At least: Does clang support the _full_ ISO C standard? Because maybe we want to compile the library with clang later...)

neithernut:
clang is supposed to be gcc-compatible, hence it should also support braces as part of an expression. But that's not my point here. Not conforming to the ISO-C standart is not acceptable imho.
The entire typesafety argument may be an issue we might consider neccessary to prevent, however, if this means _not_ doing ISO-C the solution is flawed.
I'd suggest using explicit casts without resorting to assignments. I'll push my sollution and add a PR.

neithernut:
I had a closer look at it and actually, I don't get in what way this does type checking, anyway.

neithernut:
Please have a look at PR #49.

neithernut:
Also, _please_ divide this branch into interface and implementation(s) and merge it already, so we can properly assign functions implementation to people.

matthiasbeyer:
I guess we should just restart this PR. The history is indeed a mess.

matthiasbeyer:
I'll close this now. That branch is a good example for a real fucked up history.

PR #20, opened 2014-08-27
Fix: Added indention for case-block

matthiasbeyer:
We should indent this if we need a block inside a case-statement like
so, which is _much_ more readable for long switch(-case) blocks, as we do
not visually collide with the closing bracket of the switch block.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,10 +78,11 @@ style. Here are just some examples:
     case 1:
         // do something
         break;
-    case 2: {
-        // or something else, but in a block, if you need to
-        break;
-    }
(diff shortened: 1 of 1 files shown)


TheNeikos:
makes sense

matthiasbeyer:
@neithernut , @manuelmessner opinions?

neithernut:
I have no strong feeling about this one way or another.

matthiasbeyer:
So, @TheNeikos , I can merge this?

TheNeikos:
Imo, yes

manuelmessner:
Yes, you can.

PR #23, opened 2014-08-28
Add likely.h

matthiasbeyer:
This adds the likely() and unlikely() macros for branch condition optimization.

(This is for micro-optimization. But maybe it is good to add these macros now, so one can take care of them when writing new code).

### DIFF
--- a/src/libreset/util/likely.h
+++ b/src/libreset/util/likely.h
@@ -1,0 +1,22 @@
+#ifndef __LIKELY_H__
+#define __LIKELY_H__
+
+/**
+ * @file likely.h
+ *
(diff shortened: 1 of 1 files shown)


src/libreset/util/likely.h
```diff
@@ -0,0 +1,22 @@
+#ifndef ___LIKELY_H__
```
TheNeikos: There is a typo in this guard, one underscore too much.

neithernut:
ACK

matthiasbeyer:
Rebased (fixup commit merged into appropriate commit).

PR #24, opened 2014-08-28
Add hash.h hash utility

matthiasbeyer:
This adds the `src/libreset/hash.h` file, which is a utility for dealing
with hashes.

It defines the `rs_hash` type as well as it contains functions for
dealing with objects of this type.

Simplifies #12

### DIFF
--- a/include/libreset/hash.h
+++ b/include/libreset/hash.h
@@ -5,6 +5,9 @@
 
 /**
  * The hash type
+ *
+ * The `rs_hash` type is guaranteed to be able to be handled with the normal
+ * comparators.
  */
 typedef rs_hash uint64_t;
 
--- a/include/libreset/hash.h
+++ b/include/libreset/hash.h
@@ -1,0 +1,12 @@
+#ifndef __LIBRESET_HASH_H__
+#define __LIBRESET_HASH_H__
+
+#include <stdint.h>
+
+/**
+ * The hash type
+ */
+typedef rs_hash uint64_t;
+
+#endif //__LIBRESET_HASH_H__
+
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
I guess this should go into the public interface, shouldn't it?

neithernut:
Hashes are numbers. I wouldn't make comparisions between hashes public. For internal use, we can simply use the regular numerical operations and comparisions.
All the user of libreset should have to know about hashes is what int-type we use.
Regarding the hash itself, all the user should have to supply is a function which produces a hash. Comparisions only have to be supplied for the object itself.

matthiasbeyer:
So we should the `typedef rs_hash uint64_t` into the public interface, but all other functions we implement for it into the internal stuff?

I would stick with helper functions for comparison etc., as we can change the actual implementation of the hash type later on without having to rewrite loads of code!

neithernut:
Hash types are _always_ _ever_ integral types.

matthiasbeyer:
Have a look at e1121ad please, also the commit msg.

neithernut:
Still, little hepler functions operating on the _hash_ are completely useless.

matthiasbeyer:
Updated onto latest master and appropriate to the upper discussion.

include/libreset/hash.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include <stdint.h>
+
+/**
+ * The hash type
+ *
+ * The `rs_hash` type is guaranteed to be able to be handled with the normal
+ * comparators.
+ */
+typedef rs_hash uint64_t;
```
neithernut: Would use size_t.

matthiasbeyer: Indeed a good idea.

neithernut:
ACK.

PR #25, opened 2014-08-28
Add element.h utility

matthiasbeyer:
This PR adds the `src/libreset/element.h` file, which is for dealing
with element types.

It defines the `rs_element` type as well as helper functions for
dealing with this type.

Requires #24 to be merged.
Simplifies #12.

### DIFF
--- a/src/libreset/element.h
+++ b/src/libreset/element.h
@@ -1,0 +1,87 @@
+#ifndef __ELEMENT_H__
+#define __ELEMENT_H__
+
+#include "libreset/hash.h"
+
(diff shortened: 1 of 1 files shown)


neithernut:
An AVL node will contain all the elements having identical hashes. This will result in a list/an array of elements. Having the hash as part of an element doesn't make sense here.

matthiasbeyer:
So I guess this whole PR gets rejected because of this, right?

neithernut:
Yep, I think so.

PR #27, opened 2014-08-28
Add Hashtable type

matthiasbeyer:
This PR introduces the very first idea of the hashtable type. This
type is for internal use only.

Assumes #12 to be merged.

### DIFF
--- a/src/libreset/ht.h
+++ b/src/libreset/ht.h
@@ -9,35 +9,35 @@
  * Type for hashtable bucket
  *
  */
-struct rs_ht_bucket {
+struct ht_bucket {
     struct avl avl;
 };
 
 /**
  * Hashtable type
  */
-struct rs_ht {
-    struct rs_ht_bucket* buckets;
+struct ht {
+    struct ht_bucket* buckets;
     size_t nbuckets;
 };
 
 /**
- * Initialize a struct rs_ht object
+ * Initialize a struct ht object
  *
- * @return the pointer to the passed struct rs_ht object, NULL on failure
+ * @return the pointer to the passed struct ht object, NULL on failure
  */
-struct rs_ht*
-rs_ht_init(
-    struct rs_ht* ht, //!< The hashtable object to initialize
+struct ht*
+ht_init(
+    struct ht* ht, //!< The hashtable object to initialize
     size_t n //!< The initial number of buckets
 );
 
 /**
- * Destroy a struct rs_ht object
+ * Destroy a struct ht object
  */
 void
-rs_ht_destroy(
-    struct rs_ht* ht //!< The hashtable object to destory
+ht_destroy(
+    struct ht* ht //!< The hashtable object to destory
 );
 
 /**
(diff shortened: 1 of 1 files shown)


src/libreset/ht.h
```diff
@@ -0,0 +1,78 @@
...
+#include "libreset/avl.h"
+
+/**
+ * Type for hashtable bucket
+ *
+ */
+struct rs_ht_bucket
+{
+    struct avl_el avl;
+};
```
neithernut, line 14: Having the root AVL node at the bucket structure, instead of simply a pointer inflicts some limitations/special cases in the AVL implementation.

matthiasbeyer, line 14: You mean smth like:

``` c
struct avl
{
    struct avl_el* rootnode;
};
```

? If yes, should this go into the AVL implementation or into the HT implementation? I guess it should go into the AVL implementation, right?

neithernut:
Put it into the AVL implementation.

matthiasbeyer:
Ok. I will do this today, as I'm on the train for 2h or something. As this requires changes in the whole interface of the avl implementation.

---

Please check your comment again: it is outside of the discussion path of the actual discussion about this, please ensure you comment in the right place next time! :-)

matthiasbeyer:
@neithernut can we please discuss this interface?

neithernut:
I'd prefer the type to be more opaque (hide the internal datatypes)

matthiasbeyer:
Why? This is an internal type and won't be shown to the public anyways, so why hiding the implementation?

neithernut:
Ah, I was getting confused by the functions's declarations, which I thought would be functions of the public interface.

matthiasbeyer:
Functions of the public interface only include set functionality, whereas they start with `r_set_`. So no, this is internal! :-)

matthiasbeyer:
As nobody reviews this, I started to implement it. And yes, this is meant to sound pissed-off a bit!

The changes in the `macros.h` helper should be appropriate, as the current paper sounds like these helpers are required for the implementation.

src/libreset/ht.h
```diff
@@ -0,0 +1,77 @@
...
+struct ht_bucket {
+    struct avl avl;
+};
+
+/**
+ * Hashtable type
+ */
+struct ht {
+    struct ht_bucket* buckets;
+    size_t               nbuckets
```
neithernut: please fix the formatting here.

neithernut:
To be fair, it is a bit unusual to follow things in github since the mechanism just do not work like e.g. mailinglists, where you can postpone processing of messages by marking them as unread.
Furthermore, the interface seems ok. If you could just push the header and stub implementations, a monster branch like we already have for the AVL type can be avoided.
Additionally, it's a _lot_ more clear _what_ has to be done, since single functions can be tracked through the issue system and/or exising PRs/branches.

matthiasbeyer:
I removed everything which is not interface definition.

Implementations will be added in other PRs after this PR is merged.

matthiasbeyer:
Assumes #51 to be merged now.

neithernut:
ACK.

matthiasbeyer:
#51 got merged, so I'll merge this one now.

PR #29, opened 2014-08-29
Add formatter config

manuelmessner:
Fixes issue raised in #10.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -44,9 +44,8 @@ do
     cp $file $fpath
 
     astyle --options=./etc/astylerc < $fpath > $fpath_formatted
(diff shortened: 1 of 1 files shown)


hooks/pre-commit
```diff
@@ -44,9 +44,9 @@ do
     cp $file $fpath
 
     astyle --options=./etc/astylerc < $fpath > $fpath_formatted
-    diff $fpath $fpath_formatted > /dev/null
+    sed -i '$ d' $fpath_formatted
```
matthiasbeyer, line 48: Can you explain this line (in a comment)? I'm not familiar enough with sed to exactly understand what this is doing.

PR #31, opened 2014-08-30
Updated astylerc to fit our style

matthiasbeyer:
This PR updates the `etc/astylerc` file to fit our styleguide.
#12 should be re-written, so each commit fits the style.


neithernut:
Ahhh, yes...
ACK

PR #32, opened 2014-08-30
Add doxygen config

matthiasbeyer:
This PR adds a config file for doxygen as well as the `./doc` subdir tree (via `.gitkeep` files).


neithernut:
ACK

PR #33, opened 2014-08-31
Added ./u entries in .gitignore

matthiasbeyer:
This adds two entries in the .gitignore file:

```
./u
./u/*
```

for ignoring such a subdirectory. The gentle user is now able to put
her/his utility scripts in there.

For example, one could put an astyle helper (callastyle.sh) in there:

```
#!/bin/bash
astyle --options=./etc/astyle
```

as typing the astyle call by hand can be annoying.

---

I want this patch to go into the codebase, as I can for example put vim-ex snippets in this directory and call them from vim, so they can do work for me!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,12 @@ bject files
 
 *.backup
 
+#
+# Directory where the gentle programmer can put her/his utility snippets when
+# working with the codebase.
+#
+# For example, you can put vim scripts in there when working with vim, so you
+# can load them into your vim session and do work for you.
+#
+./u
+./u/*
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
In addition, this wouldn't hurt us that much, as I don't think we will have a subdirectory `u` in the root directory of the project, will we?

neithernut:
nope, but we should document this feature somewhere (so that users who use utility-script will "mkdir u" and put their scripts into ./u)
Maybe in contibuting.md

neithernut:
ACK.

PR #34, opened 2014-08-31
Fix: We do not do kernel patches here, generalize the statements

matthiasbeyer:
This was a copy-and-paste failure.


neithernut:
ACK.

PR #35, opened 2014-09-01
Add linkedlist

matthiasbeyer:
This adds an implementation for a linked list.

@neithernut please review. @manuelmessner and @TheNeikos, you are welcome, too!

### DIFF
--- a/src/libreset/ll.c
+++ b/src/libreset/ll.c
@@ -2,11 +2,6 @@
 
 #include <stdlib.h>
 
-struct ll*
-ll_alloc(void) {
-    return calloc(1, sizeof(struct ll));
-}
-
 void
 ll_destroy(
     struct ll* ll
--- a/src/libreset/ll.c
+++ b/src/libreset/ll.c
@@ -11,6 +11,14 @@ void
 ll_destroy(
     struct ll* ll
 ) {
+    struct ll_element* iter = ll->head;
+    struct ll_element* next;
+
+    while (iter) {
+        next = iter->next;
+        free(iter);
+        iter = next;
+    }
 }
 
 struct ll*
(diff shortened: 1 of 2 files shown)


src/libreset/ll.h
```diff
@@ -0,0 +1,43 @@
...
+    struct ll_element*  next;
+    void*               data;
+};
+
+struct ll {
+    struct ll_element* head;
+};
+
+struct ll*
+ll_alloc(void);
```
neithernut: We won't need an allocator. Just put a `struct ll` into the node struct.

src/libreset/ll.c
```diff
@@ -0,0 +1,90 @@
...
+
+    for (iter = ll->head; iter; iter = iter->next) {
+        if (iter == del) {
+            last->next = iter->next;
+            free(iter);
+            break;
+        }
+
+        last = iter;
+    }
```
neithernut, line 65: This does not work if the first element is deleted.

src/libreset/ll.c
```diff
@@ -0,0 +1,89 @@
...
+    struct ll* ll,
+    struct ll_element* del
+) {
+    struct ll_element* iter;
+    struct ll_element* last;
+
+    for (iter = ll->head; iter; iter = iter->next) {
+        if (iter == del) {
+            if (iter == ll->head) {
+                ll->head = iter->next;
```
neithernut: You check every element for being the head element?
Why don't you just check whether you delete the first element before running into the loop?

matthiasbeyer: That was indeed stupid.

neithernut:
ACK.

PR #36, opened 2014-09-02
Remove me from the list of authors

matthiasbeyer:
I don't think I should be listed as author of the paper anymore, @neithernut writes the paper!


neithernut:
For the sake of progress: ACK.

PR #41, opened 2014-09-07
Added .travis.yml file

matthiasbeyer:
This PR adds the `.travis.yml` file and closes #18 therefor.

PR #44, opened 2014-09-07
Use old method for setting C-Flags

neithernut:
This PR modifies how CMake will set the C-Flags for the build.


matthiasbeyer:
Hm. I cannot test whether this solves the issues in #41 . I would like this to be a seperate PR, but also to check before merging.

(For documentation:) What I will do: Check if the cmake setup still works (locally), if yes merge and rebase #41 .

matthiasbeyer:
Seems to work.

Merged.

PR #46, opened 2014-09-07
Convert comments

matthiasbeyer:
This PR converts the comments to be ISO-C90 compatible.

### DIFF
--- a/include/libreset/hash.h
+++ b/include/libreset/hash.h
@@ -11,5 +11,5 @@
  */
 typedef rs_hash size_t;
 
-#endif //__LIBRESET_HASH_H__
+#endif /* __LIBRESET_HASH_H__ */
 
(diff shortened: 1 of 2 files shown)


neithernut:
ISO-C90? Why bother? Every reasonable compiler knows (at-least) C99.

matthiasbeyer:
Because for me, the current build setup prints a number of warnings because of this comment style!

This is an approach to fix the code, an approach for fixing this with the build setup would be appropriate, too. But I like the comment style much more, it is more C-like, I think.

But well, that's personal taste...

neithernut:
Fixing the build-setup is, indeed more reasonable. Await a PR from me

matthiasbeyer:
Ok. I will close this unmerged.

PR #49, opened 2014-09-09
Add scopeless macros

neithernut:
This makes the macros `MIN(x,y)` and `MAX(x,y)` apply to ISO-C.
It also "fixes" the type checking by casting both arguments to the first argument's type.

### DIFF
--- a/src/libreset/util/macros.h
+++ b/src/libreset/util/macros.h
@@ -37,10 +37,9 @@
  * @note Opens own scope, so the temp variables do not show up outside of the
  * macro.
  */
-#define MIN(x,y)                    \
-       ({ __typeof__ (x) _x = (x);  \
-          __typeof__ (y) _y = (y);  \
-          _x < _y ? _x : _y; })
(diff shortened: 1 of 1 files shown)


src/libreset/util/macros.h
```diff
@@ -24,10 +24,9 @@
  * @note Opens own scope, so the temp variables do not show up outside of the
  * macro.
  */
-#define MAX(x,y)                    \
-       ({ __typeof__ (x) _x = (x);  \
-          __typeof__ (y) _y = (y);  \
-          _x > _y ? _x : _y; })
+#define MAX(x,y)                                    \
+        ((__typeof__(x)) x > (__typeof__(x)) y ?    \
```
matthiasbeyer, line 28: Will this result in a compiler warning if `typeof(x) != typeof(y)` ?

neithernut, line 28: Not, is they are castable into another. However, the original version would not generate a useful message either.

PR #51, opened 2014-09-10
Add util macros

matthiasbeyer:
This PR was created from patches from the #27 PR. We seperate them here, so we do not get this annoying long history for every type we implement.

### DIFF
--- a/src/libreset/util/macros.h
+++ b/src/libreset/util/macros.h
@@ -41,6 +41,14 @@
         ((__typeof__(x)) x > (__typeof__(x)) y ?    \
          (__typeof__(x)) y : (__typeof__(x)) x)
 
+/**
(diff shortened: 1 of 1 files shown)


neithernut:
Not sure if we need LSB and MSB. However, those may be removed afterwards.
Conclusion: ACK.
Note: this may result in a merge conflict with PR #49.

matthiasbeyer:
This was updated. The first commit was removed in the rebase. Please review again:

---

Note: You don't have to mention potential merge conflicts. I **always** wait for the merge check github does. If it says that there are merge conflicts, I rebase or notify the submitter to rebase.

matthiasbeyer:
As there was only stuff _removed_ from the history, but not added, I assume the ACK from @neithernut as still valid.

PR #52, opened 2014-09-10
Added AVL header (interface for avl type)

matthiasbeyer:
This is related to #12 .
#12 became a mess, so I want to re-start the PR by opening this one, including _only_ the interface definition for the AVL type.

Please review.

---

I would close #12 when we use this instead!

### DIFF
--- a/src/libreset/avl.h
+++ b/src/libreset/avl.h
@@ -1,0 +1,163 @@
+/**
+ * @file avl.h
+ *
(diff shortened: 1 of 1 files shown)


neithernut:
ACK.

matthiasbeyer:
Shall I close #12 therefor? I would open several issues for implementing the functions then...

Issue #16, opened 2014-08-27
Missing: Testing


matthiasbeyer:
Tests should be added. We can start this from now, I guess, as parts of the interface are already defined.

neithernut:
I do not consider the interface stable atm.

matthiasbeyer:
Ok. I'll keep this open anyways, as it is a TODO.

matthiasbeyer:
I will split up this one into several issues now and close this one therefor.

PR #58, opened 2014-09-10
[WIP] Add testing for linked list implementation

matthiasbeyer:
This PR adds tests for the linked list implementation.

### DIFF
--- a/tests/ll/ll_test.c
+++ b/tests/ll/ll_test.c
@@ -29,6 +29,26 @@ START_TEST (test_ll_delete_data) {
 }
 END_TEST
 
+START_TEST (test_ll_insert_multiple) {
+    struct ll* ll = malloc(sizeof(*ll));
+    int i;
+    int data[] = {
+        0, 1, 2, 3, 4, 5, 6, 7, 8, 9,
+    };
+
+    for (i = 0; i < 10; i++) {
+        ck_assert_ptr_eq(ll_insert_data(ll, &(data[i])), ll);
+        ck_assert_ptr_eq(ll->head->data, &(data[i]));
+    }
+
+    while (ll->head) {
+        ck_assert_ptr_eq(ll_delete(ll, ll->head), ll);
+    }
+
+    ll_destroy(ll);
+}
+END_TEST
+
 Suite*
 suite_ll_create(void) {
     Suite* s;
@@ -40,6 +60,7 @@ suite_ll_create(void) {
     case_insert = tcase_create("Inserting");
     tcase_add_test(case_insert, test_ll_insert_data);
     tcase_add_test(case_insert, test_ll_delete_data);
+    tcase_add_test(case_insert, test_ll_insert_multiple);
 
     /* Adding test cases to suite */
     suite_add_tcase(s, case_insert);
--- a/tests/ll/ll_test.c
+++ b/tests/ll/ll_test.c
@@ -37,12 +37,12 @@ START_TEST (test_ll_insert_multiple) {
     };
 
     for (i = 0; i < 10; i++) {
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
@neithernut Can you please add the build setup.

We use the `check` testing framework here.

neithernut:
Yep, I'll have a look at it. You may open another issue for that.
The test might be expanded btw.
ACK for now.

matthiasbeyer:
I'll not merge this without build setup and I would prefer the build setup patches go into this PR. Too much PRs and Issues create too much noise in the repository, somewhere we have to keep things short and compact.

neithernut:
k, I'll add a note on how to include this test after I've brought up the infrastructure.

matthiasbeyer:
Any updates on this? I would like to write the tests as soon as possible!

neithernut:
I focused on the paper since the milestone is due today. However, I'll assign more time to doing this PR's prequisites. Hopefully, I'll be done with it in 24h max.

neithernut:
I will need to build _something_, else the PR will not be very meaningul.
I recommend merging at least `tests/tests.c` with the one test stripped. After that is done I can push a setup.
(As a temporary work-around I started a private integration branch on this machine)

matthiasbeyer:
I will create a PR for the basic testing infrastructure.

matthiasbeyer:
@neithernut Thanks for your testing build setup stuff, this seems to work now as expected. I will add more tests for the LL implementation from now on (in this PR).

matthiasbeyer:
@neithernut btw, there is a commit in this PR which fixes the build main file as well. I will not put this into a seperate PR, as this would again cause noise in the repo which is kind-of unneccessary.

neithernut:
Why did you discard the NULL terminated list of tests?

matthiasbeyer:
Because it was buggy somehow.

matthiasbeyer:
Any suggestions what I also could test?

matthiasbeyer:
@neithernut Can you have a look at the travis log and interpret that message for me?

neithernut:
Some function does not exist: `undefined reference to 'ck_assert_ptr_eq'`

matthiasbeyer:
For documentation:

On Ubuntu (travis-ci ubuntu: 12.04) only version 0.9.8 if "check" is available. A possible fix _could_ be installing another version from the backports, but it _seems_ that it is not available in the backports of Ubuntu 12.04 (I'm not an Shitbuntu guy, so I don't know how to search them properly...).

---

I'm not happy with the fact that the build servers at travis use Ubuntu, but I cannot change it. So we try to live with the fact here. It is against our philosophy, but hey ... we need a CI and travis is easy to setup and configure when it comes to the `travis.yml` thing. We should try to avoid trouble with the OS, so we can focus on important things without having this bullshit from Canonical in our way.

---

Anyways, using functionality from pre-v0.9.8 and v0.9.8. of check should be enough.

matthiasbeyer:
Relates to #16 btw.

matthiasbeyer:
I'm a bit confused that there is _so_ little test output. I mean... why do I name the tests? I assumed at least some output like "Test foobar: sucess" or something.

@neithernut can you please review _really_ carefully, also the output from travis, if the tests actually ran?

---

edit: I did, too, but I'm insecure about this one :-P

neithernut:
It sais `Running suite(s): Linkedlist`, so i assume it will list all the tests it's running.
I also assume that it will list all the tests that failed.
Maybe there's some option you can enable to have more output?

matthiasbeyer:
Ahh, yes! I will add the appropriate flag.

neithernut:
Now that's a little more output.

neithernut:
ACK for merge.

Issue #54, opened 2014-09-10
Implement: avl_destroy()


matthiasbeyer:
Implement: `avl_destroy()`

neithernut:
May now be closed.

PR #65, opened 2014-09-10
Implemented avl_destroy()

matthiasbeyer:
Solves #54 .


src/libreset/avl.c
```diff
@@ -0,0 +1,28 @@
+#include <stdlib.h>
+
+#include "libreset/avl.h"
+
+static void
+destroy_node(
```
neithernut: As this function does not destroy a single node but an entire subtree, I'd rather name this function `destroy_subtree` or something else, but not `destroy_node`.

neithernut:
ACK.

PR #66, opened 2014-09-10
 [WIP] Implement avl add

matthiasbeyer:
Solves #55 .

### DIFF
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -87,6 +87,11 @@ insert_element_into_tree(
     struct avl_el** root //!< The root element of the tree where to insert
 );
 
+static struct avl_el*
+element_with_hash(
+    struct avl_el* root,    //!< The element where to begin to search in
+    rs_hash h               //!< The hash to search for
+);
 
 /*
  *
@@ -173,6 +178,37 @@ insert_element_into_tree(
     struct avl_el* el, //!< The element to insert
     struct avl_el** root //!< The root element of the tree where to insert
 ) {
+}
+
+/**
+ * Find an element by hash
+ *
+ * @return Ptr to the found element or NULL if there is none.
+ */
+static struct avl_el*
+element_with_hash(
+    struct avl_el* root,    //!< The element where to begin to search in
+    rs_hash h               //!< The hash to search for
+) {
+    if (!root) {
+        return NULL;
+    }
+
+    if (root->hash == h) {
+        return root;
+    }
+
+    if (root->hash >= h) {
+        if (root->l) {
+            return element_with_hash(root->l, h);
+        }
+    } else {
+        if (root->r) {
+            return element_with_hash(root->r, h);
+        }
+    }
+
+    return NULL;
 }
 
 /*
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -81,6 +81,12 @@ new_avl_el(
     rs_hash h //!< The hash for the new struct avl_el object
 );
 
+static struct avl_el*
+insert_element_into_tree(
+    struct avl_el* el, //!< The element to insert
+    struct avl_el** root //!< The root element of the tree where to insert
+);
+
 
 /*
  *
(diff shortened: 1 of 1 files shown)


neithernut:
Acceptable.
ACK.

matthiasbeyer:
You would leave the rotation functions unimplemented for this PR?

neithernut:
The interface for the rotation functions should not change. Hence, this PR can be ACK'ed without the rotation functions being merged already.
We may, however, defer the merge until the rotation functions are merged.

matthiasbeyer:
So, I'll add another issue for the rotation functions

src/libreset/avl.c
```diff
@@ -7,6 +11,170 @@ avl_alloc(void) {
...
+    }
+
+    if (unlikely(el->hash == (*root)->hash)) {
+        cmp = 0;
+    } else {
+        cmp = (el->hash > (*root)->hash) ? -1 : 1;
+    }
+
+    if (unlikely(cmp == 0)) {
+        /* Shouldn't be possible */
```
TheNeikos: If this is impossible, how come that it might get set to 0 earlier? Why not check directly at this point whether
`unlikely(el->hash == (*root)->hash)` is true and drop the one before?

src/libreset/avl.c
```diff
@@ -7,6 +11,170 @@ avl_alloc(void) {
...
+        if (avl_height((*root)->r) - avl_height((*root)->l) == 2) {
+            if (likely(el->hash != (*root)->r->hash)) {
+                *root = single_rotate_with_right(*root);
+            } else {
+                *root = double_rotate_with_right(*root);
+            }
+        }
+    }
+
+    (*root)->height = MAX(avl_height((*root)->l), avl_height((*root)->l)) + 1;
```
TheNeikos: Is it correct to check for Max between `x` and `x` in this case?

src/libreset/avl.c
matthiasbeyer: Of course, I should simplify this algorithm here!

src/libreset/avl.c
matthiasbeyer: Nope, this was an typo, thank you for mentioning!

neithernut:
The rotation functions will be introduced in a seperate PR, so please remove the last commit, see PR #72. Also note that the structure of avl.c is changed by that PR.
One the rotation functions are introduced, I'd use those in the insertion and deletion functions.

matthiasbeyer:
@neithernut The PR still mentiones some issues because the rotation functions are neither defined nor implemented. Shall I wait until they are written or can I merge this right now and you'll add the implementations?

neithernut:
The PR bringing the rotation functions are on the way. See PR #74.

matthiasbeyer:
waiting on #74 .

neithernut:
Not that #74 is done, you can easily adapt your approach:
- In `insert_element_into_tree()`, you call `regen_metadata()` on the current node on your way "up" (after insertion that is). You can remove the rotations from that function. They are not needed anymore.
- In `add()` you simply call `rebalance_subtree()`, after adding a node and be done with it. The tree will be automatically rebalanced.

matthiasbeyer:
I will continue working on this whole stuff when I'm healthy again! Thanks for the explanation, though!

neithernut:
Ok. get better!

src/libreset/avl.c
```diff
@@ -117,6 +178,91 @@ avl_find(
...
+        return *root;
+    }
+
+    if (el->hash < (*root)->hash) {
+        (*root)->l = insert_element_into_tree(el, &(*root)->l);
+    } else {
+        (*root)->r = insert_element_into_tree(el, &(*root)->r);
+    }
+
+    (*root)->height = MAX(avl_height((*root)->l), avl_height((*root)->r)) + 1;
```
neithernut: regen_metadata does this for you, already.

src/libreset/avl.c
```diff
@@ -117,6 +178,89 @@ avl_find(
...
+) {
+    if (!root) {
+        return NULL;
+    }
+
+    if (root->hash == h) {
+        return root;
+    }
+
+    if (root->hash >= h) {
```
neithernut: I'd like to see the recursion resolved the same way it's in `avl_find`, as we do not have to handle anything after recusring.
Site note: another PR may then reimplement `avl_find` by the means of this function.

neithernut:
Nice, I completely overlooked that you can simply use `avl_find` anyway.
ACK for squash and merge.

matthiasbeyer:
\m/ finally! I'll do the squashing right now!

PR #68, opened 2014-09-10
Add tests astyle

neithernut:
This PR introduces astyle tests. The tests are run using the CTest framework. After running `cmake .`, simply run `ctest .` in the same directory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ before_install:
 
 install:
     - sudo apt-get install -q
+        astyle
         gcc-4.8
         texlive-metapost
         texlive-latex-base
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,4 @@ script:
     - cmake .
     - make
     - make paper
+    - ctest -VV .
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,9 @@ CMakeFiles
 Makefile
 *.cmake
 
+# Test related stuff
+Testing
+
 #
 # Directory where the gentle programmer can put her/his utility snippets when
 # working with the codebase.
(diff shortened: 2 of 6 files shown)


neithernut:
Note: the pre-commit hook may be rewritten and based on ctest  after this is merged. @manuelmessner: would you like to do this?

neithernut:
Meh... again some packages not working as they do here... I'll fix this after dinner.

neithernut:
@matthiasbeyer: please have a look at the last section in the Travis CI log. Seems like the version of astyle availible (2.01) doesn't know about a few options:
"style", "allign-reference", "max-code-length", "break-after-logical"
Should we ignore them based on the asyle version installed?

Note: I'm acutaly quite pissed about how aastyle. It misbehaves, adds lines, is not forward compatible and does not have an option for simply checking a file's style. I'd like to introduce an alternative as soon as one becomes availible.

matthiasbeyer:
How does this thing work? Is `ctest` a cmake builtin?

@neithernut you are welcome to replace the style checker (in another PR). It should (at least) support the features astyle has, of course. And it should be available to all common platforms/distros (Arch, Gentoo, Debian and the family, fedora and the family and the suse family).

.travis.yml
```diff
@@ -9,6 +9,7 @@ before_install:
 
 install:
     - sudo apt-get install -q
+        astyle
```
matthiasbeyer, line 12: Well, if this installs a wrong version, you may can install a version from the backports somehow?

But as you said, if astyle behaves bad (and it does), we may use another tool.

neithernut:
ctest comes with cmake

neithernut:
Waiting for issue #69.

neithernut:
I'll close this PR. If issue #69 is resolved, a new PR will be pushed.
Some of the commits will be salvaged for tests done with the check framework.

Issue #56, opened 2014-09-10
Implement: avl_del()


matthiasbeyer:
Implement: `avl_del()`

PR #71, opened 2014-09-11
Implement avl del

matthiasbeyer:
This PR implementes the `avl_del()` function.

Some linkedlist helpers were introduced.

**Note:** Some possibly needed rotations are still missing.

### DIFF
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -147,8 +147,7 @@ swp_avl_els(
 static void
 delete_node_from_subtree(
     struct avl_el* node, //!< The node to delete
-    struct avl_el* parent, //!< The parent to delete
-    struct avl_el* root //!< The root node of the subtree
+    struct avl_el* parent //!< The parent to delete
 );
 
 /*
@@ -266,7 +265,7 @@ avl_del(
         return avl->root;
     }
 
-    delete_node_from_subtree(found, parent, avl->root);
+    delete_node_from_subtree(found, parent);
 
     rebalance_subtree(avl->root);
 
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
Sovles #56

matthiasbeyer:
Waiting on #74 as well as on patches which add the rotation-calls in the delete algorithm.

neithernut:
I added a note to PR #66, which explains how to use the functions introduced in PR #74.

neithernut:
Uh... you'll probaly need a recursive aproach here to accomodate for regen_metadata to be called for each node from the point of deletion up to the root node.
Also: maybe stuff can be split up.
Also: instead of copying a node's contens, relinking would be better.

neithernut:
@matthiasbeyer: if you're interested, I can write some helper functions, e.g. for swapping/picking elements. You would then use those for the `avl_del()` function.

matthiasbeyer:
I am interested. But please in an own PR.

neithernut:
I'll do so.

src/libreset/avl.c
```diff
@@ -203,6 +284,122 @@ insert_element_into_tree(
...
+    if (node->hash != root->hash) {
+        if (node->hash < root->hash) {
+            delete_node_from_subtree(node, parent, root->l);
+        } else {
+            delete_node_from_subtree(node, parent, root->r);
+        }
+        rebalance_subtree(root);
+    } else {
+
+        /*
```
matthiasbeyer, line 399: @neithernut If I understand things right, I can replace this whole `else` branch with the `isolate_root` function now, right?

neithernut, line 399: Yep.

---

However, I'd just recurse into each subtree looking for nodes satisfying the predicate and delete them. You may also call `rebalance_subtree()` directly in `avl_del()`.

neithernut, line 399: Point is: looking for nodes satisfying a predicate and then looking for the same nodes _Again_, now based on their hashes, is pointless.

matthiasbeyer, line 399: There's a differance between hashes and predicate: A hash can identify several elements, a predicate cannot. If the predicates evaluates to true, the element is removed and after that returned!

neithernut, line 399: Yes, but you are, at that moment, already _sitting_ on the node in question. Why not perform the deletion right there, if the node is empty?

neithernut, line 399: And how can a predicate not hte true for a number of elements?

matthiasbeyer, line 399: Not "hte" true?

With my comment, I meant the `else` block... not the recursion above. But that's true, I can delete this part.

matthiasbeyer, line 399: An predicate can be false because there are _several_ elements per _node_. And we only delete _one element_ not _one node_!

neithernut, line 399: Yep, but the predicate could be true for several elements which sit on possibly several nodes.

matthiasbeyer, line 399: That complicates a lot of things. I wouldn't do that!

neithernut, line 399: In contrary, is _simplyfies_ a lot of things.

matthiasbeyer, line 399: I'd welcome your patches on this (and everything else related to this PR) :+1:

matthiasbeyer:
I guess we can throw this one away due to #111

neithernut:
Yep.

PR #72, opened 2014-09-12
Fix avl implementation and build

neithernut:
This PR is the first of a row of PRs which will ultimately introduce internal rotation functions.
This PR fixes a couple of issues.
This also teaches me one thing: have a closer look at the PRs I ack.


neithernut:
This PR will probably conflict with both PR #66 and PR #71. Those will have to be rebased, one this is merged.

PR #73, opened 2014-09-12
Add node cnt

neithernut:
This PR adds the number of nodes to the node's metadata.
Storing the number of nodes of a subtree will effectively speed up some operations.

### DIFF
--- a/src/libreset/avl.h
+++ b/src/libreset/avl.h
@@ -41,10 +41,10 @@ struct avl {
  */
 struct avl_el {
     rs_hash hash;
-    struct ll ll;       //!< The linkedlist where the data elements are stored
-    signed int height;  //!< The height of the subtree
-    struct  avl_el* l;  //!< Next left node
-    struct  avl_el* r;  //!< Next right node
+    struct ll ll;           //!< The linkedlist where the data elements are stored
+    unsigned int height;    //!< The height of the subtree
+    struct  avl_el* l;      //!< Next left node
+    struct  avl_el* r;      //!< Next right node
 };
 
 /**
--- a/src/libreset/avl.h
+++ b/src/libreset/avl.h
@@ -139,6 +139,17 @@ avl_height(
 );
 
 
+/**
+ * Get the number of nodes a subtree contains
+ *
+ * @return The number of nodes the subtree contains
+ */
+inline static unsigned int
+avl_node_cnt(
+    struct avl_el* root //!< The root element
+);
+
+
 /*
  *
  *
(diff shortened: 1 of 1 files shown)


neithernut:
Note: the inlined functions `avl_height()` and `avl_node_cnt()` may be malicious in some cases.
The trinary operator (`?:`) may evaluate the second and third operand _before_ looking at the condition. This may result in a segfault.
Fixing this may be done in another PR. I'd suggest making them _not_ inline but aim for usage of LTO. This is feasible as the code byse will be quite small compared to other libraries.

src/libreset/avl.h
```diff
@@ -41,10 +41,11 @@ struct avl {
  */
 struct avl_el {
     rs_hash hash;
-    struct ll ll;       //!< The linkedlist where the data elements are stored
-    signed int height;  //!< The height of the subtree
```
matthiasbeyer, line 45: `height` was wrong here. But that's why it was `signed` - it actually was the balance level, which is in `[-1, 0, 1]` and should therefor be `signed`.

src/libreset/avl.h
```diff
@@ -151,13 +163,20 @@ avl_get_hash(struct avl_el* el) {
...
     struct avl_el* root //!< The root element
 ) {
     return root == NULL ? 0 : root->height;
 }
 
+inline static unsigned int
+avl_node_cnt(
+    struct avl_el* root //!< The root element
+) {
+    return root == NULL ? 0 : root->node_cnt;
```
matthiasbeyer: I'd prefer parentheses about the condition here. I think this improves the readability MUCH more than it adds noise to the code.

matthiasbeyer:
The changes look good, but let me be sure about the naming:

When speaking about _nodes_, you mean the nodes in the tree, not the _elements_ in the tree. I consider _elements_ as values which are stored, which can be greater than the number of _nodes_. Do we agree on that?

Just to be sure about the things we talk, so we do not mess things up by stumbling about the naming of things!

neithernut:
Yep, nodes are the nodes of the tree. The number of elements in the subtree may be greater.

src/libreset/avl.h
neithernut, line 45: You can't propagate the balance level upwards (to the root node) in a trivial way. Having the real (maximum) height would be the better choice. It's more easily propagatable and, together with the node-count, you can derive whether _all_ the subtrees within a subtree are balanced or not.

src/libreset/avl.h
neithernut: We will move the implementation into the source file anyway. I will, however, add parantheses.

src/libreset/avl.h
matthiasbeyer, line 45: Ok. So ACK from me for this part.

neithernut:
@matthiasbeyer: please review again and merge.

Issue #57, opened 2014-09-10
Implement: avl_find()


matthiasbeyer:
Implement: `avl_find()`

neithernut:
May be closed now.

PR #75, opened 2014-09-12
Implement avl find

matthiasbeyer:
Solves #57 

**Note:** We return a `struct avl_el*` here, so we do not search for an actual _element_ but a _node_ where the element we look for should be in.

### DIFF
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -41,6 +41,23 @@ avl_destroy(
     return 1;
 }
 
+struct avl_el*
+avl_find(
+    struct avl* avl,
+    rs_hash hash
+) {
+    struct avl_el* iter = avl->root;
+
+    while (iter && iter->hash != hash) {
+        if (iter->hash > hash) {
+            iter = iter->l;
+        } else {
+            iter = iter->r;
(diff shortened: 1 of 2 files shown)


src/libreset/avl.c
```diff
@@ -41,6 +41,27 @@ avl_destroy(
...
+
+    while (iter && iter->hash != hash) {
+        if (iter->hash > hash) {
+            iter = iter->l;
+        } else {
+            iter = iter->r;
+        }
+    }
+
+    if (iter && iter->hash != hash) {
```
neithernut: I don't see how you can ever get an `iter` which is not `NULL` but which has a different hash for what you're looking for.
Hence, the entire if may be omitted.

matthiasbeyer: Sure, I messed up again... _tadada_

neithernut: Well, It just does what it's supposed to do, so you're fine. It's just an unnecessary step.

neithernut:
ACK.

matthiasbeyer:
rebased. Waiting on CI for merge.

Issue #61, opened 2014-09-10
Implement: ht_find()


matthiasbeyer:
Implement: `ht_find()`

PR #82, opened 2014-09-13
Implement ht find

matthiasbeyer:
Solves #61 

Note: I assume `s` from the paper being the size (number of buckets) of the hashtable.

### DIFF
--- a/src/libreset/ht.c
+++ b/src/libreset/ht.c
@@ -38,6 +38,6 @@ ht_find(
         return NULL;
     }
 
-    return &ht->buckets[hash % (CONSTPOW_TWO(BITCOUNT(hash)) / ht->nbuckets)];
+    return &ht->buckets[hash / ((CONSTPOW_TWO(BITCOUNT(hash)) / ht->nbuckets))];
 }
 
--- a/src/libreset/ht.c
+++ b/src/libreset/ht.c
@@ -21,8 +21,9 @@ ht_destroy(
     struct ht* ht
 ) {
     if (ht) {
-        for (; ht->nbuckets > 0; ht->nbuckets--) {
-            avl_destroy(&ht->buckets[ht->nbuckets - 1].avl);
+        size_t i = ht_nbuckets(ht);
+        while (i--) {
+            avl_destroy(&ht->buckets[i].avl);
         }
         free(ht->buckets);
         free(ht);
--- a/src/libreset/ht.c
+++ b/src/libreset/ht.c
@@ -1,6 +1,7 @@
 #include <stdlib.h>
 
 #include "ht.h"
+#include "util/macros.h"
 
 struct ht*
 ht_init(
@@ -27,3 +28,16 @@ ht_destroy(
         free(ht);
     }
 }
+
+struct ht_bucket*
+ht_find(
+    struct ht* ht,
+    rs_hash hash
+) {
+    if (!ht) {
+        return NULL;
+    }
+
+    return &ht->buckets[hash % (CONSTPOW_TWO(BITCOUNT(hash)) / ht->nbuckets)];
+}
+
--- a/src/libreset/ht.c
+++ b/src/libreset/ht.c
@@ -39,6 +39,6 @@ ht_find(
         return NULL;
     }
 
-    return &ht->buckets[hash / ((CONSTPOW_TWO(BITCOUNT(hash)) / ht->nbuckets))];
+    return &ht->buckets[hash >> (BITCOUNT(hash) - ht->sizeexp)];
 }
 
--- a/src/libreset/ht.c
+++ b/src/libreset/ht.c
@@ -9,8 +9,8 @@ ht_init(
     size_t n
 ) {
     if (ht) {
-        ht->buckets = calloc(n, sizeof(*ht->buckets));
-        ht->nbuckets = n;
+        ht->buckets = calloc(CONSTPOW_TWO(n), sizeof(*ht->buckets));
+        ht->sizeexp = n;
     }
 
     return ht;
--- a/src/libreset/ht.h
+++ b/src/libreset/ht.h
@@ -45,9 +45,11 @@ struct ht_bucket {
 
 /**
  * Hashtable type
+ *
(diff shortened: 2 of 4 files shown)


src/libreset/ht.c
```diff
@@ -0,0 +1,15 @@
...
+struct ht_bucket*
+ht_find(
+    struct ht* ht,
+    rs_hash hash
+) {
+    if (!ht) {
+        return NULL;
+    }
+
+    return &ht->buckets[hash % (CONSTPOW_TWO(BITCOUNT(hash)) / ht->nbuckets)];
```
neithernut: Have a look at the paper. The bucket is _not_ selected by applying the modulo, but by "selecting" the most significant bits of the hash.

matthiasbeyer: I already read this part of the paper, but I guess I do not understand the approach. Your patches are welcome!

matthiasbeyer: Actually, this line is a implementation of fig 6 from the paper, so I do not know what's wrong here!

neithernut: The equation goes `h(i) / (2^N/s)`, not `h(i) % (2^N/s)`! That's a division, not a modulo.

matthiasbeyer: Actually not. The paper says modulo.

matthiasbeyer: OMFG. This was a reading error. The character for division is the same as for modulo, in some way.

We should definitely fix this in the paper and use a colon there!

matthiasbeyer:
@neithernut Can you tell what this error means (from the build):

```
/home/travis/build/matthiasbeyer/libreset/src/libreset/ht.c: In function ‘ht_find’:
/home/travis/build/matthiasbeyer/libreset/src/libreset/ht.c:41:5: error: left shift count >= width of type [-Werror]
     return &ht->buckets[hash / ((CONSTPOW_TWO(BITCOUNT(hash)) / ht->nbuckets))];
```

and how to fix it?

---

By the way: You wrote in the paper that we use the most significant **bits**, but you did not state how many bits...

neithernut:
The error means that you shift so many bits to the left that the resulting value will always be `0`.

---

You must only use so many bits as there are neccessary to hold the buckets' indices. The term given in the paper should yield just that.

matthiasbeyer:
Then, I'm not capable of this, as the implemented function literally _is_ the term from the paper.

neithernut:
However, you run into an issue regarding the number of bits you have here. Hence, you have to calculate the value a bit differently.

matthiasbeyer:
"Fixing" the macros with casts to `size_t` doesn't help.

neithernut:
Yep, the forula has to be adjusted instead.
Since the number of buckets is always `2^x`, you may just store `x` in the ht data type. You can then just do something like:

``` c
bucket[hash >> (width_of_hash - x)]
```

neithernut:
You could also calculate `x` on the fly, but why not just put it into the ht.

neithernut:
Btw: this issue is a perfect example of how CI and -Werror help to remove bugs before they are introduced into the master/dev branch!

matthiasbeyer:
Putting it into the `struct ht` type and replacing the `nbuckets` would cause more changes: The `_init` and `_destroy` function make use of the `nbuckets` member. This can be changed, of course, but then we would have to calculate the number of buckets there...

Of course, this would be much more efficient, as `_init` and `_destroy` are called just once for a `struct ht` object.

I will try to implement it this way.

neithernut:
The information density is much higher if you just store the exponent.
Also, you can get `nbuckets` by doing a simple left-shift. Doing the opposite takes a few lines.

src/libreset/ht.h
```diff
@@ -46,10 +46,12 @@ struct ht_bucket {
...
 /**
  * Hashtable type
+ *
+ * The type for the hashtable, holding buckets and size.
  */
 struct ht {
-    struct ht_bucket* buckets;
-    size_t nbuckets;
+    struct ht_bucket* buckets; //!< The buckets of the hashtable
+    size_t sizeexp; //!< Exponent 2 must be raised to, to get the size of the ht
```
neithernut: please add a comma or something after "Exponent"

src/libreset/ht.c
```diff
@@ -20,10 +21,24 @@ ht_destroy(
     struct ht* ht
 ) {
     if (ht) {
-        for (; ht->nbuckets > 0; ht->nbuckets--) {
-            avl_destroy(&ht->buckets[ht->nbuckets - 1].avl);
+        size_t i = ht_nbuckets(ht);
+        for (; i > 0; i--) {
+            avl_destroy(&ht->buckets[i - 1].avl);
```
neithernut: Declaring an iterator before the loop and then having a 0-check as the condition seem odd.
Imho the following would be more elegant:

``` c
size_t i = ht_nbuckets(ht);
while (--i) {
    avl_destroy(&ht->buckets[i].avl);
}
```

src/libreset/ht.c
```diff
@@ -20,10 +21,24 @@ ht_destroy(
     struct ht* ht
 ) {
     if (ht) {
```
neithernut, line 23: Early return? Would be more elegant imho.

matthiasbeyer, line 23: Nah, I don't care about it if there are less then 10 lines in the if branch.

``` c
if (foo) {
    bar();
    baz();
}
```

or

``` c
if (!foo) {
    return;
}
bar();
baz();
```

doesn't make that much difference.

src/libreset/ht.c
```diff
@@ -20,10 +21,24 @@ ht_destroy(
...
-        for (; ht->nbuckets > 0; ht->nbuckets--) {
-            avl_destroy(&ht->buckets[ht->nbuckets - 1].avl);
+        size_t i = ht_nbuckets(ht);
+        while (--i) {
```
neithernut: I made a mistake earlier. It should be `i--` rather than `--i`.

neithernut:
ACK for squashing and merge.

PR #83, opened 2014-09-13
Applied astyle on macros.h

matthiasbeyer:
Somehow this file was missed on the last run and gets in my way much too often when committing. So I'll add this PR to format it. Finally.


neithernut:
ACK.

Issue #79, opened 2014-09-13
Add bloom filter


neithernut:
We will use bloom filters in our datastructure to speed things up. The paper will contain a section about bloom filter usage and the AVL will be extended to make use of bloom filters.
This issue may be used to track the progress on this.

PR #88, opened 2014-09-14
Add bloom filter

neithernut:
This PR adds the bloom filter data type and functions.
It also contains adjustments to the AVL data structure.

### DIFF
--- a/src/libreset/bloom.c
+++ b/src/libreset/bloom.c
@@ -21,6 +21,8 @@
  */
 #include "bloom.h"
 
+#include "params.h"
+
 /*
  *
  *
@@ -33,5 +35,31 @@
  * Number of bits in a bloom filter
  */
 #define BLOOM_BITS (sizeof(bloom)*8)
+
+
+/*
+ *
+ *
+ * interface implementation
+ *
+ *
+ */
+
+bloom
+bloom_from_hash(
+    rs_hash hash
+) {
+    bloom result = 0;
+
+    unsigned int vars = HASH_VARIANTS;
+    while (vars--) {
+        result |= 1 << (hash % BLOOM_BITS);
+
+        // transform the hash to generate new variant
+        hash /= BLOOM_BITS;
+    }
+
+    return result;
+}
 
 
--- a/src/libreset/bloom.c
+++ b/src/libreset/bloom.c
@@ -86,6 +86,14 @@ bloom_from_hash(
 }
 
 int
+bloom_may_contain(
+    bloom element,
+    bloom set
+) {
+    return (element & (~set)) == 0;
+}
+
+int
 bloom_may_have_common(
     bloom a,
     bloom b
--- a/src/libreset/bloom.c
+++ b/src/libreset/bloom.c
@@ -1,0 +1,24 @@
+/*
+ * libreset - Reentrent set library for fast set operations in C
+ *
+ * Copyright (C) 2014 Matthias Beyer
+ * Copyright (C) 2014 Julian Ganz
+ *
+ * This file is part of libreset.
+ *
+ * libreset is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * libreset is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with libreset. If not, see <http://www.gnu.org/licenses/>.
+ */
+#include "bloom.h"
+
+
--- a/src/libreset/bloom.c
+++ b/src/libreset/bloom.c
@@ -21,4 +21,17 @@
  */
 #include "bloom.h"
 
+/*
+ *
+ *
+ * definitions
+ *
+ *
+ */
+
+/**
+ * Number of bits in a bloom filter
+ */
+#define BLOOM_BITS (sizeof(bloom)*8)
+
 
--- a/src/libreset/bloom.c
+++ b/src/libreset/bloom.c
@@ -40,6 +40,29 @@
 /*
  *
  *
+ * forward declarations
+ *
+ *
+ */
+
+/**
+ * Special function only to be used in bloom_may_have_common()
+ *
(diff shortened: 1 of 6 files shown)


src/libreset/bloom.c
```diff
@@ -0,0 +1,151 @@
...
+int
+bloom_may_contain(
+    bloom element,
+    bloom set
+) {
+    return (element & (~set)) == 0;
+}
+
+int
+bloom_may_have_common(
```
neithernut, line 97: Note that this algorithm may seem a bit strange, but given the right compiler optimisation, it should be much faster than a classical approach for calculating the number of bits in an integer.

src/libreset/bloom.c
```diff
@@ -0,0 +1,151 @@
...
+) {
+    bloom result = 0;
+
+    unsigned int vars = HASH_VARIANTS;
+    while (vars) {
+        result |= 1 << (hash % BLOOM_BITS);
+
+        // transform the hash to generate new variant
+        hash /= BLOOM_BITS;
+    }
```
matthiasbeyer, line 83: By now, this is a `while(true)` loop!

src/libreset/bloom.c
```diff
@@ -0,0 +1,151 @@
...
+    bloom a,
+    bloom b
+) {
+    // calculate the number of bits common to both filters
+    bloom section = a & b;
+
+    // calculate the number of bits in "section"
+
+    // to do this, we will consider "section" to be made up by ints of single
+    // bits. Those will be added to ints of two bits, four bits and so on.
```
matthiasbeyer: This is a perfect case why I prefer using non-C99 comment style. It just looks ugly this way.

src/libreset/bloom.c
```diff
@@ -0,0 +1,151 @@
...
+ *
+ */
+
+
+bloom
+erasure_mask(
+    unsigned int width
+) {
+    // we select the 2nd integer, starting from the least significant
+    bloom mask = ((1 << width) - 1) << width;
```
matthiasbeyer: Maybe you should add a cast here: `((size_t) 1) << width`.

src/libreset/bloom.h
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Bloom filter type
+ *
+ * The bloom filter is a probabilistic set representation used for speedups.
+ */
+typedef size_t bloom;
+
+/**
+ * Get the bloom filter for an element with a given hash.
+ *
```
matthiasbeyer, line 50: Missing: `@memberof bloom` and note that this is a "static" function (when speaking of OOP-static).

src/libreset/bloom.h
```diff
@@ -0,0 +1,86 @@
...
+__const__ bloom
+bloom_from_hash(
+    rs_hash hash //!< hash value for which to calculate the bloom filter
+);
+
+/**
+ * Check whether an element may be in a bloom filter
+ *
+ * This is much faster than checking whether two sets may have common elements
+ *
```
matthiasbeyer, line 63: Also here is the `@memberof bloom` missing.

src/libreset/bloom.h
```diff
@@ -0,0 +1,86 @@
...
+ * @return 0 if the element is not in the set, 0 otherwise
```
matthiasbeyer: returns zero or zero? Typo here?

src/libreset/params.h
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+#ifndef __PARAMS_H__
+#define __PARAMS_H__
+
+/**
+ * Number of variants to derive from a hash function
+ *
+ * To be used for the bloom filter
+ */
```
matthiasbeyer, line 40: Can you please provide more verbose documentation and explanation on this one?

Also: Is this a header for stuff which may be overridden by users? If yes, this should probably go into the public interface, shouldn't it?

src/libreset/bloom.c
neithernut, line 83: Meh, forgot something.
Will fix.

src/libreset/bloom.c
neithernut: I agree, and I usually would not comment this way. However, I do not think that most people would get how it works easily without such extensive comments.

matthiasbeyer: I do not get how this works either. I meant you should prefer `/* */` comment style over `//` comment style.

src/libreset/params.h
neithernut, line 40: Nope, this will _not_ be overridden by users.

src/libreset/bloom.c
neithernut: Nope, I'd rather keep comments short.
If I were to explain this algorithm in the source code, you'd have a perfectly dense wall of text. `/* */` in function/implementation source code just motivates you to do just that!

src/libreset/params.h
neithernut, line 40: In fact, it will probably be derived from some constants, like `sizeof(size_t)`

src/libreset/bloom.c
neithernut, line 83: Fixed.

src/libreset/bloom.c
neithernut: I'll try to break it down, so it's really only one-liners

src/libreset/bloom.h
neithernut, line 50: Should be fixed now.

src/libreset/bloom.h
neithernut, line 63: Fixed.

src/libreset/bloom.c
neithernut: I rephrased most of the comments. Should be a bit clearer now.

src/libreset/bloom.c
neithernut, line 97: Is there a need to explain the algo (e.g. in the paper)?

neithernut:
Those will be squashed anyways

src/libreset/bloom.c
matthiasbeyer, line 97: Of course.

At least because your comment confuses me even more. The number of bits in an integer is constant and can be retreived with `sizeof(int)<<3` - so your comment doesn't make any sense to me.

neithernut, line 97: An integer is nothing but an array of bits of arbitrary length representing a number.
I'll add a section to the paper...

neithernut:
Do I have a GO for squashing?

matthiasbeyer:
Yes for squashing. I don't know if I will merge, as I guess there should be more documentation around.

neithernut:
Squashed and rebased.

neithernut:
Fixes issue #79.

PR #89, opened 2014-09-14
Add tests infrastructure

neithernut:
This PR adds very basic infrastucture which will serve as a prequisite for a variety of tests.

PR #91, opened 2014-09-14
Added test main file

matthiasbeyer:
This PR adds a main file for the testing setup, containing test execution.

### DIFF
--- a/tests/tests.c
+++ b/tests/tests.c
@@ -1,0 +1,36 @@
+#include <check.h>
+
+/*
+ * Include test files below
+ */
+
+typedef Suite* (*suite_creator_f)(void);
+
+int
+main(void) {
+    int nfailed = 0;
+
+    Suite* s;
+    SRunner* sr;
+
+    suite_creator_f iter;
+
+    /*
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Relates #58 , #89

neithernut:
ACK, obviously.

matthiasbeyer:
mhm... I don't know if this compiles or anything... but yeah, that's obvious that you cannot the setup before setting it up completely.

So I'll merge.

matthiasbeyer:
Obviously, this was not reviewed. But now it is in the master.

There are at least two problems with the code which result in a compile-error. We are in pre-1.0, but this stuff really should not happen! Not even in pre-1.0!

neithernut:
In this case, it's a bootstrapping problem. Someone _should_ have tried to build the file by hand, but that's a missed opportunity now.
You may either push a fix or wait for me to fix it as soon as it becomes neccessary.

matthiasbeyer:
The latter should be enough.

neithernut:
Done, see PR #89.

PR #98, opened 2014-09-16
Fix: We should use several test cases here

matthiasbeyer:
Fixup for bad review (me too) of the linkedlist testing stuff.


neithernut:
ACK.

Issue #62, opened 2014-09-10
Implement: ht_insert()


matthiasbeyer:
Implement: `ht_insert()`

matthiasbeyer:
Seems like @TheNeikos is on this.

neithernut:
May be closed.

Issue #103, opened 2014-09-16
Adapt to Layer model


neithernut:
The library is divided into a few layers implicitly.
The AVL uses the linked-list, the hashtable uses the AVL.
I'd like to change a few things to do the same in code. E.g. the hashtable should not fiddle with the linked list. When inserting an element, the hashtable should insert the element into the AVL, which in turn will possibly add new nodes on it's own. The AVL will then add the element to the linked list, which should check for duplicates by iterating and applying the cmp-function provided by the user.
Requires: type information provided by the user being passed down the layers.

matthiasbeyer:
That's already the case, isn't it?

When I do a `avl_add()` if finds the appropriate node (or creates one) and adds the element to the linkedlist inside it.

neithernut:
It doesn't check whether the element exists in the linked list, which should be done in the linked list. To do this, the type information has to be passed.
`avl_del` should be able to destroy elements using the user provided type information, too.

neithernut:
Also, `avl_find` returns a node, which requires the ht to fiddle around with the linked list.

matthiasbeyer:
On the first point: That's a simple ptr-compare, you don't need type information for this! And yes, this should be done by the linked list, this is indeed a thing which _could_ be improved.

The implementation for `avl_find()` should change, that's true! Are you willing to do this?

neithernut:
The pointer being identical is _not_ the only way, two elements might be identical. E.g. consider `strcmp()`.
a PR will be submitted probably this afternoon/evening.

matthiasbeyer:
You think people will do things like this:

```
int a = 4;
rs_insert(set, (char*)&a);
rs_insert(set, ((char*)&a)+1);
```

or what? Because this is the only thing I can imagine one could insert an "object" twice without the same ptr.

Storing the type information (assuming you mean size here) too, would prevent us from cases like this with little overhead. But it would also cause changes, _and I think we're running out of time for library_.

neithernut:
This is neccessary for the library to function correctly. This is how a set works.

neithernut:
It actually shouldn't be hard, since `r_set_cfg` already exists.

matthiasbeyer:
indeed. Awaiting PR on this.

neithernut:
Servicing info: I'm working on adaption of the AVL interface, so you guys can adapt the ht stuff.

matthiasbeyer:
Lol. 5 merged PRs. closing this one now.

PR #104, opened 2014-09-16
Add ht insert tests/implementation

TheNeikos:
This is meant to solve #62.

I am not sure if the test I added is sufficient.

### DIFF
--- a/tests/ht/ht_test.c
+++ b/tests/ht/ht_test.c
@@ -1,0 +1,28 @@
+#include <check.h>
+
+#include <stdlib.h>
+#include "ht.h"
+
+START_TEST (test_ht_insert_data) {
+}
+END_TEST
+
+
+Suite*
+suite_ht_create(void) {
+    Suite* s;
+    TCase* case_insert;
+
+    s = suite_create("Hashtable");
+
+    /* Test case creation */
+    case_insert         = tcase_create("Inserting");
+
+    /* test adding */
+    tcase_add_test(case_insert, test_ll_insert_data);
+
+    /* Adding test cases to suite */
+    suite_add_tcase(s, case_insert);
+
+    return s;
+}
--- a/tests/ht/ht_test.c
+++ b/tests/ht/ht_test.c
@@ -4,6 +4,13 @@
 #include "ht.h"
 
 START_TEST (test_ht_insert_data) {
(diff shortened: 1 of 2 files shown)


tests/ht/ht_test.c
```diff
@@ -0,0 +1,35 @@
+#include <check.h>
+
+#include <stdlib.h>
+#include "ht.h"
+
+START_TEST (test_ht_insert_data) {
+    struct ht* ht = malloc(sizeof(*ht));
+    ht_init(ht, 4);
+    int data = 5;
+    ck_assert(ht_insert(ht, &data, sizeof(data)));
```
matthiasbeyer, line 10: I'd like you to compare against something here, please add a compare to `1` or what the expected return value is. Use `ck_assert_int_eq` if this is appropriate!

TheNeikos, line 10: Doing that now.

matthiasbeyer:
Besides the one point ths looks good so far.

For a TDD workflow, you should add patches to pass the test now (on this very same PR). I'd like to suggest more tests, too: Adding multiple values, adding multiple values which collide in their hashes, etc.

Edit: The title should be adapted, then.

tests/ht/ht_test.c
matthiasbeyer, line 10: Nah, the interface of the ht lacks here.

You need to pass the hash of the element, too. The hashtable should not create the hash of the object (@neithernut right?).

neithernut, line 10: It should, using a user provided function, see issue #103.

matthiasbeyer, line 10: So we need interface changes to be able to add a hash function.

TheNeikos:
I will wait for #103 and start from there again. Seems to be the most sensible to me.

neithernut:
You can probably adapt quickly once the changes are done, so keep the branch.

matthiasbeyer:
@TheNeikos #104 changes the interface of `ht_insert()`.

matthiasbeyer:
Please resubmit onto master as soon as the appropriate interface changes are done. My attempt to switch to kind-of multi-branch-workflow was bullshit, we will submit PRs onto master until v0.0.1, please apologize the confusion!

PR #105, opened 2014-09-16
[WIP] Fix ht interface

matthiasbeyer:
This PR is for fixing the hashtable interface. **I also changes the public interface.**

PR #106, opened 2014-09-16
Mod avl interface

neithernut:
This PR will adapt the AVL interface to the layer structure laid out in issue #103.

PR #107, opened 2014-09-16
[WIP] Fix ht interface

matthiasbeyer:
Resubmit from #105 onto master

### DIFF
--- a/src/libreset/ht.h
+++ b/src/libreset/ht.h
@@ -101,7 +101,6 @@ rs_hash
 ht_insert(
     struct ht* ht, //!< The hashtable object to insert into
     void* data, //!< The data
-    size_t datasize //!< The size of the data in bytes
 );
 
 /**
--- a/src/libreset/ht.h
+++ b/src/libreset/ht.h
@@ -87,7 +87,9 @@ ht_destroy(
 struct ht_bucket*
 ht_find(
     struct ht* ht, //!< The hashtable object to search in
-    rs_hash hash //!< The hash of the element to find
+    rs_hash hash, //!< The hash of the element to find
+    void* cmp, //!< Element to compare against
+    struct r_set_cfg* cfg //!< type information provided by user
 );
 
 /**
--- a/src/libreset/ht.h
+++ b/src/libreset/ht.h
@@ -103,6 +103,7 @@ rs_hash
 ht_insert(
     struct ht* ht, //!< The hashtable object to insert into
     void* data, //!< The data
+    struct r_set_cfg* cfg //!< type information provided by user
 );
 
 /**
--- a/src/libreset/ht.h
+++ b/src/libreset/ht.h
@@ -113,9 +113,9 @@ ht_insert(
  *
  * @memberof ht
  *
- * @return the deleted element or NULL if not found
+ * @return number of removed elements
  */
-void*
+unsigned int
 ht_del(
     struct ht* ht, //!< The hashtable object to delete from
     rs_hash hash, //!< The hash of the element to delete
--- a/src/libreset/ht.h
+++ b/src/libreset/ht.h
@@ -95,9 +95,9 @@ ht_find(
  *
  * @memberof ht
  *
- * @return 1 on success, else zero
+ * @return the hash of the inserted element
  */
-int
+rs_hash
 ht_insert(
     struct ht* ht, //!< The hashtable object to insert into
     void* data, //!< The data
--- a/src/libreset/ht.h
+++ b/src/libreset/ht.h
@@ -112,10 +112,12 @@ ht_insert(
  *
  * @return the deleted element or NULL if not found
  */
-struct element*
+void*
 ht_del(
     struct ht* ht, //!< The hashtable object to delete from
-    rs_hash hash
+    rs_hash hash, //!< The hash of the element to delete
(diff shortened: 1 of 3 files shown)


neithernut:
ACK.

matthiasbeyer:
waiting on CI.

matthiasbeyer:
CI failed because of stuff which should be fixed by #106 , waiting on this one.

matthiasbeyer:
@neithernut Please review again, I removed `ht_find()` as the interface for this function completely changed. I will file an issue for reimplementing this one asap this PR is merged.

src/libreset/ht.h
```diff
@@ -95,13 +99,13 @@ ht_find(
  *
  * @memberof ht
  *
- * @return 1 on success, else zero
+ * @return the hash of the inserted element
  */
-int
+rs_hash
```
neithernut: Why return the hash? This way the function doesn't tell the user, whether the insertion was successful or not.

matthiasbeyer: But the user has to know the hash of the inserted element so he can retreive it afterwards.

src/libreset/ht.h
```diff
@@ -111,10 +115,12 @@ ht_insert(
  *
  * @return the deleted element or NULL if not found
  */
-struct element*
+void*
 ht_del(
     struct ht* ht, //!< The hashtable object to delete from
-    rs_hash hash
+    rs_hash hash, //!< The hash of the element to delete
+    void* cmp, //!< Element to compare against
```
neithernut, line 122: Maybe this function should use the same semantics `avl_del` uses now: take a predicate, instead of cmp and hash, and return the number of element removed. This way we can easily emulate the old behaviour while gaining new possibilities. However, this also slows things down because every bucket has to be accessed.
Maybe we should have two functions: one to remove elements based on a predicate and one based on a comparable. We could adapt the avl interface to have both versions, too. (open an issue then)

src/libreset/ht.h
neithernut: The hash is nowhere sufficient. The user may use a key of some soft, used by the hash function, to retrieve elements: the user may create a dummy object (a "comparable") and feed it to `ht_find`. That function will then search an object which will compare to the one provided and return that one to the user.
Also, the hash may be generated by the user anyways, since _he_ or _she_ provides the hash function in the first place.

matthiasbeyer: You start to confuse me again. When inserting something, you get back a value to identify your object with. You have to pass this value to retreive the object from the library again.

So the hash is _relevant_ even if it is _not_ sufficient.

src/libreset/ht.h
matthiasbeyer, line 122: See #108 - I would like to keep the `_del` functions.

neithernut, line 122: Agreed.

src/libreset/ht.h
neithernut: The user knows the internal structure of the object and is therefore better equipped to extract a key of some sort for identification of an object. In the end, the hash is nothing but a way to speed things up.

matthiasbeyer: So you would like me to return a status and not the hash. Okay. So when the user inserts a value, her/his provided hash function is used for the hashtable, but the user will not get the hash, because  he can recreate it easily, am I right?

We should _definitely_ state in the docs that the hash is not supposed to be a hash in the known form.

neithernut: What's a hash in the known form?
It is _completely_ common for hash table implementations to require the user to provide a hash function. See the C++ STL, or uthash, or ...
Hash functions may be provided for known standard types, but _no one_ would take the binary data of an object and compute a hash from that data, if the object may possibly not be a POD.

matthiasbeyer: By "known form" I mean something like SHA or so, checksum hashes.

I never said that it is not common for libraries to ask the user for a hash function!

neithernut: The underlying hash function may still be MD5 or SHA, but it is expected to be applied to only a part of the object and that the user will do so in the function he or she provides.

neithernut:
Relates to issue #103.

src/libreset/ht.h
```diff
@@ -109,12 +113,14 @@ ht_insert(
  *
  * @memberof ht
  *
- * @return the deleted element or NULL if not found
+ * @return number of removed elements
  */
-struct element*
+unsigned int
```
neithernut, line 118: Why not `int`?

matthiasbeyer, line 118: because `int` _can_ be negative, and a counter is always positive. This way it is clear that a negative return value is not possible.

neithernut:
Also: please rebase. This may resolve the issue of the CI failing.

src/libreset/ht.h
neithernut, line 118: `ht_del()` _always_ deletes _at most_ one element. So returning whether the element was removed should suffice.

matthiasbeyer, line 118: I thought we agreed on one function removing elements with similar hashes and an `_ndel` function removing also elements which have not the same hash, as discussed in #108

neithernut:
Ahhh, the CI fails because you renamed `hashf`.

neithernut:
ACK.

Issue #109, opened 2014-09-16
Fix avl_del


neithernut:
Adapt `avl_del` to comply with PR #108.

matthiasbeyer:
are you doing this right now?

neithernut:
Nope, feel free to start a PR.

neithernut:
Just modify the interface, I'll take care of the internals as soon as I finished the LL interface.

matthiasbeyer:
I'll do this now.

matthiasbeyer:
@neithernut is this issue only for the interface, which gets solved by #111 ? Or is it also meant for the implementation and should be kept open after #111 is merged?

neithernut:
The implementation is covered by #63, so this issue may now be closed.

Issue #108, opened 2014-09-16
Functions for deleting non-hash-equivalent elements


matthiasbeyer:
As suggested by @neithernut  in #107 , we maybe should have functions for deleting elements with non-equivalent hashes as well.

I'd like to suggest:
- Keep the current functions, `_del`, let them remove (multiple) elements with the same hash only
- Add functions `_ndel` for deleting multiple elements which may not have the same hash

neithernut:
Agreed.

matthiasbeyer:
Relevant issues:

#144 , #134 , #136

neithernut:
May now be closed.

PR #111, opened 2014-09-16
Adapt avl del

matthiasbeyer:
Solves #109 , relates #108 .

### DIFF
--- a/src/libreset/avl.h
+++ b/src/libreset/avl.h
@@ -106,6 +106,7 @@ unsigned int
 avl_del(
     struct avl* avl, //!< The avl where to search in
     rs_hash hash, //!< hash value associated with d
+    void* cmp, //!< element to compare against
     struct r_set_cfg* cfg //!< type information provided by the user
 );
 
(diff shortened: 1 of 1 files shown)


neithernut:
We also need a `void* data` argument to identify the item to delete in the linked list.

matthiasbeyer:
Isn't this covered by the predicate already?

neithernut:
What predicate? you removed the predicate.
Anyways, the whole point of having `_del` and `_ndel` was to have a specialized, fast function for removal of _one_ element, fow which you have to provide a comparable, and a function to remove any element for which a predicate is true.

neithernut:
Relates to issue #103.

matthiasbeyer:
I meant the _predicate_ as the _cfg_, sorry for messing these two up here.

neithernut:
In this case it's _not_ covered by the predicate.

matthiasbeyer:
Ah, now I understand: You meant a comparator with the `void*`, which is passed to the comparator function as second argument? I will add this.

neithernut:
ACK.

matthiasbeyer:
Rebased. Waiting on CI.

PR #112, opened 2014-09-16
Added avl_ndel() interface

matthiasbeyer:
Solves #110 .

### DIFF
--- a/src/libreset/avl.h
+++ b/src/libreset/avl.h
@@ -122,6 +122,22 @@ avl_del(
 );
 
 /**
+ * Delete elements from the avl tree by predicate
+ *
+ * This function will remove elements from the avl tree which match the
+ * predicate defined in the configuration `cfg`.
+ *
+ * @memberof avl
+ *
+ * @reutrn the number of removed elements
+ */
+unsigned int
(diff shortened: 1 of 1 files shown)


neithernut:
I would add a warning about functions not accepting `NULL`, but this may also be done in the implementation PR.
Otherwise: ACK

matthiasbeyer:
Travis failes. We should fix it before merging!

src/libreset/avl.h
```diff
@@ -122,6 +122,22 @@ avl_del(
...
+ *
+ * @memberof avl
+ *
+ * @reutrn the number of removed elements
+ */
+unsigned int
+avl_ndel(
+    struct avl* el, //!< The avl where to search in
+    struct r_set_cfg* cfg //!< type information provided by the user
+};
```
neithernut: The build fails because of the typo in this line. It should be a `)`, not a `}`.

neithernut:
Relates to issue #103.

neithernut:
As already pointed out: ACK

matthiasbeyer:
Waiting on CI.

Issue #114, opened 2014-09-16
Go through the interfaces which are stable atm and add __attributes__


neithernut:
E.g. all the functions containing warnings about NULL pointers resulting in crashes may be attributed with `nonnull`.
The resulting compiler errors should also be fixed.
This will save us a lot of pain.

neithernut:
Note: for doing this looking at the source code of the signle functions should be enough.
Also: this will probably solve the issue surfacing in PR #99 as well.

matthiasbeyer:
We should probably do this in pair-programming!

neithernut:
Probably a good idea.

matthiasbeyer:
Do you think we can already start this? We could meet up today and do this, but if there are interface changes left to do, we should wait until they are done!

neithernut:
We can do it for the linked list and the AVL. Those interfaces should be stable by now. Which means: additions as soon as we introduce the fancier features, but no changes to existing function declarations.

matthiasbeyer:
Okay, meet me at our IRC for planning when to meet!

neithernut:
Mh... `avl_ndel()` is messed up, will fix. After that, the interface _really_ is stable.

neithernut:
This issue was split up into #138 and #139.

PR #127, opened 2014-09-17
Added alias for nonnull argument attribute macro

matthiasbeyer:
Adds `nonnull` macro helper.


matthiasbeyer:
@neithernut , @manuelmessner , @TheNeikos Not WIP anymore, the latest patches fixes the issue we experienced here.

---

The problem was:
We defined the macros and included `stdlib.h` afterwards. We probably wrote some stuff in `stdlib.h` which caused the bug. Reverting the order of includes in the `ll.c` file fixes the issue for me.

---

Please review and ack.

neithernut:
ACK.
After the merge, issue #114 may be addressed.

PR #130, opened 2014-09-17
Added a test for checking for duplicates in linked list

TheNeikos:
This adds a test for linked lists to assure that no duplicates can be inserted.


neithernut:
ACK.

Issue #139, opened 2014-09-18
Add __attributes__ to function interface for AVL implementation


matthiasbeyer:
This one is for me.

PR #140, opened 2014-09-18
Add function attrs avl

matthiasbeyer:
Solves #139 and therefor #114 partly.

### DIFF
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -123,7 +123,9 @@ __attribute__((nonnull (1)))
 static void
 regen_metadata(
     struct avl_el* node //!< The node to regenerate
-);
+)
+__attribute__((nonnull (1)))
+;
 
 /**
  * Create a new struct avl_el object
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -38,7 +38,9 @@ __attribute__((nonnull (1, 2)))
 static struct avl_el*
 rebalance_subtree(
     struct avl_el* root //!< The root of the subtree to rebalance
-);
+)
+__attribute__((nonnull (1)))
+;
 
 /**
  * Rotate a node counter-clockwise
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -149,7 +149,9 @@ static struct avl_el*
 insert_element_into_tree(
     struct avl_el* el, //!< The element to insert
     struct avl_el** root //!< The root element of the tree where to insert
-);
+)
+__attribute__((nonnull (1, 2)))
+;
 
 /**
  * Find a node by it's key/hash
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -162,7 +162,9 @@ static struct avl_el*
 find_node(
     struct avl* avl,
     rs_hash hash
-);
+)
+__attribute__((nonnull (1)))
+;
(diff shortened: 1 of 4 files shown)


src/libreset/avl.h
```diff
@@ -62,7 +62,9 @@ struct avl_el {
  * @return A pointer to a new avl_el object or NULL on failure
  */
 struct avl*
-avl_alloc(void);
+avl_alloc(void)
+__warn_unused_result__
+;
```
neithernut, line 67: The malloc attribute would also make sense here.

matthiasbeyer, line 67: I was not sure about this, so I left it. But I'll add it.

matthiasbeyer:
The `__nonnull__` does not work here. So I used the plain notation with `__attribute__`.

`__nonnull__` will be removed by upcoming patches in this PR.

src/libreset/avl.h
matthiasbeyer, line 67: Nah, the `avl_alloc()` function does not make sense, because it is embedded into another data structure. We should remove it.

src/libreset/avl.h
```diff
@@ -153,7 +166,9 @@ avl_find(
 static inline rs_hash
 avl_get_hash(
     struct avl_el* el //!< The element to get the hash for
-);
+)
+__deprecated__
```
neithernut, line 170: This function is used internally, so either the function itself or the functionality should be moved to the source file.

neithernut:
What about all the functions in `avl.c`?

src/libreset/avl.h
matthiasbeyer, line 170: So `__deprecated__` is just fine here! :-)

matthiasbeyer:
What should be with them? The attributes to their interface definition should be sufficient, shouldn't it?

neithernut:
E.g. the entire point of having the nonnull attribute is to have compiler warnings/errors on errorneous usage. You generally also want them for internal funtions.

matthiasbeyer:
that's right. I'll add these as soon as I'm back.

neithernut:
One of the tests has to be adjusted (see the Travis CI output or locally run `ctest -VV .`).
This is a perfect example of why to have this.

neithernut:
The offending function call is in `ht.c`, please fix.

matthiasbeyer:
@neithernut (or @TheNeikos ) feel free to play around with the latest commit. It re-adds the use of the `__nonnull__` alias to the PR. For me, this breaks the build. I guess travis will also fail.

src/libreset/avl.c
```diff
@@ -6,8 +6,8 @@
 #include "util/macros.h"
```
neithernut, line 6: Multiple inclusions of `util/macros.h`?
Also, `avl.h` _should_ be the first inclusion in the file.
If this does not work, we need some more work on `attributes.h` to cope with other (e.g. standart) headers.
I'll probably look into this tomorrow.

matthiasbeyer, line 6: :+1:

neithernut:
As issue #224 is fixed, this may now be rebased and reworked.

matthiasbeyer:
Cannot rebase that easy. The PR is much too old to "just rebase" it. I'll close this one.

Issue #143, opened 2014-09-18
Where is the ll_is_empty() function?


matthiasbeyer:
Where is the ll_is_empty() function?

Just wondered where it went, because I need it for the implementation of `avl_ndel()`.

neithernut:
Seems it doesn't exist. I'll implement it and submit a PR right away.

matthiasbeyer:
:+1: 

Seems it was removed by someone. Hm, strange!

neithernut:
I'm away in about 30min. If you merge PR #141, I can rebase and resolve merge conflicts before being afk.

PR #146, opened 2014-09-18
Add ll_is_empty

neithernut:
This PR solves issue #143.


neithernut:
Before merging: I'll add a little change to an avl function before...

neithernut:
Done. Ready for merge.

matthiasbeyer:
The code for this function could be implemented right in the header and marked as `inline`. But well, I don't care much about this at this moment.

Waiting for CI.

neithernut:
With this small library we can afford putting _all_ the implementation into the source files.
Hail LTO! (still has to be activated, though...)

PR #133, opened 2014-09-18
[WIP] Add avl test

TheNeikos:
This adds a few more tests for the AVL

PR #147, opened 2014-09-18
Expand ll tests

matthiasbeyer:
Expands the ll tests by inserting more assertions and adding a test for the new empty checker function.

### DIFF
--- a/tests/ll/ll_test.c
+++ b/tests/ll/ll_test.c
@@ -18,10 +18,9 @@ START_TEST (test_ll_insert_data) {
     int data = 5;
     ll_insert(ll, &data, &cfg_int);
 
-    ck_assert(ll->head->next == NULL);
-    ck_assert(ll->head->data == &data);
+    ck_assert_int_eq(1, ll_delete(ll, &data, &cfg_int));
+    ck_assert_int_eq(0, ll_delete(ll, &data, &cfg_int));
 
-    free(ll->head);
     free(ll);
 }
 END_TEST
--- a/tests/ll/ll_test.c
+++ b/tests/ll/ll_test.c
@@ -18,6 +18,7 @@ START_TEST (test_ll_insert_data) {
     int data = 5;
     ll_insert(ll, &data, &cfg_int);
 
+    ck_assert(ll->head->next == NULL);
     ck_assert(ll->head->data == &data);
 
     free(ll->head);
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
@TheNeikos please review and ack.

neithernut:
I'm not a friend on fiddeling with internas of data structures.
The tests should _only_ use the interface of a datastructure!

matthiasbeyer:
But how do you test the interface of the datastructure when only using the interface itself?

neithernut:
By observing the result of operations.
Example: if you insert an item successfully, it should be able to delete it. If you try to delete the same item twice or an item which is not in the list, the function should return false.
Also: checking for internas is bad testing because
- internas may change
- you tend to write tests based on implementation details, possibly lacking other aspects.

Have a look at what @TheNeikos did in the AVL tests (PR #133).

matthiasbeyer:
I fully understand your point. But how do you guarantee that the deletion algorithms works here?

neithernut:
E.g. try to find the element. If you found it _before_ the deletion and you do not find it _after_ the deletion, `ll_del()` behaves as expected.

TheNeikos:
I agree with that.

A test called `ll_find_test` can test beyond that, but should focus on find. It is crucial to make sure that everything works together! That way I cannot change `ll_del` _without_ making sure it always works as intended. With TDD you spec out your interface (and implicitely the usage) so that you can be sure it actually looks right beforehand.

matthiasbeyer:
Travis failed with:

```
2: /home/travis/build/matthiasbeyer/libreset/tests/ll/ll_test.c:21:F:Inserting:test_ll_insert_data:0: Assertion '1==ll_delete(ll, &data, &cfg_int)' failed: 1==1, ll_delete(ll, &data, &cfg_int)==1
```

Can someone please explain this to me? Why the hell does it fail?

neithernut:
Works on this machine...
Can you somehow trigger a rebuild?

matthiasbeyer:
@neithernut started rebuild. Travis seems to be under heavy load atm, so this could take a while...

neithernut:
So now I have _two_ machines on which this test passes...
Something _must_ be wrong with the CI.

matthiasbeyer:
@neithernut , @TheNeikos seems to be solved by the latest commit. The problem was (as for #156) the `ck_assert_int_eq()` macro.

neithernut:
May now be rebased and merged.

matthiasbeyer:
No need for rebase here, merge?

neithernut:
Yep.

Issue #144, opened 2014-09-18
Implement ll_ndel()


matthiasbeyer:
Implement ll_ndel()

neithernut:
I'll take this one. But I'll perform the changes sometime tonight/tomorrow.

matthiasbeyer:
Ok. I need this for `avl_ndel` which is required for `ht_ndel`. So this blocks a lot of things.

matthiasbeyer:
@neithernut Sorry, could not wait.

neithernut:
May now be closed.

PR #145, opened 2014-09-18
[WIP] Implementation of avl_ndel()

matthiasbeyer:
Added implementation of avl_ndel() with helper function, which is not implemented at this moment, as #143 and #144 are not done yet, but required for this implementation.

PR #148, opened 2014-09-18
Implement ll_ndel()

matthiasbeyer:
Solves #144 .

### DIFF
--- a/src/libreset/ll.c
+++ b/src/libreset/ll.c
@@ -100,6 +100,41 @@ ll_delete(
     return 0;
 }
 
+unsigned int
+ll_ndel(
+    struct ll* ll,
+    r_predf pred,
+    void* etc,
+    struct r_set_cfg* cfg
+) {
+    unsigned int cnt = 0;
+
+    struct ll_element** p_next = &ll->head;
+    struct ll_element* iter = ll->head;
+    struct ll_element* next;
+
+    while (iter) {
+        next = iter->next;
+        if (pred(iter->data, etc)) {
+            *p_next = next;
+
+            if (cfg->freef) {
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
Required for #145

src/libreset/ll.c
```diff
@@ -100,6 +100,41 @@ ll_delete(
...
+    struct ll* ll,
+    r_predf pred,
+    void* etc,
+    struct r_set_cfg* cfg
+) {
+    unsigned int cnt = 0;
+
+    struct ll_element** p_next = &ll->head;
+    struct ll_element* iter = ll->head;
+    struct ll_element* next;
```
neithernut: Having `iter`, `next` and `p_next` doesn't make sense. The code can easily be rewritten to use only `p_next` or an equivalent iterator and a temporary fariable for the case of deletion. See `ll_del()`.

matthiasbeyer: `ll_del()` is buggy the way it is written. This code works. You're welcome to send me a PR (onto this) to simplify this.

neithernut: How is `ll_delete()` buggy?

matthiasbeyer: Nah, buggy is the wrong term here, I meant that it targets another problem, because with `ll_ndel()` you can delete multiple entries, you must be careful because there could be the possibility you have to delete `n` elements following each other.

Writing the code like in `ll_delete()` just does not work here.

neithernut: Oh yes, it does, see PR #155.

matthiasbeyer:
Assuming this is ACKed by you, @neithernut , as you did the last commit. Merging now.

Issue #92, opened 2014-09-14
Add extensive debug printing


matthiasbeyer:
#90 introduced a header for debug printing.

Use this header and add extensive debug printing.

@manuelmessner , @theneikos One of you wanna do this?

matthiasbeyer:
I will not close this until v0.0.1 is ready and the debug printing is sufficient in my opinion.

matthiasbeyer:
Well, afer overthinking my last statement, I guess this can be closed, too. We already have _a lot_ of debug printing, I don't think there is need for more by now.

PR #152, opened 2014-09-18
Add debug printing for ll

matthiasbeyer:
Partly solves #92

### DIFF
--- a/src/libreset/ll.c
+++ b/src/libreset/ll.c
@@ -24,6 +24,16 @@
 
 #include "ll.h"
 
+#include "util/debug.h"
+
+/**
+ * Debug print helper for linkedlist implementation code
+ *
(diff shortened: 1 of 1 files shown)


src/libreset/ll.h
```diff
@@ -33,6 +33,7 @@
 #define __LL_H__
 
 #include "util/attributes.h"
+#include "util/debug.h"
```
neithernut: Not needed.

matthiasbeyer: Why not? Sure this is needed for the `dbg()` macro!

matthiasbeyer: Ah, it was the header,...

neithernut:
ACK.

matthiasbeyer:
Waiting on CI after rebase.

PR #158, opened 2014-09-19
Implement avl_ndel()

neithernut:
This PR completes PR #145.

### DIFF
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -568,8 +568,6 @@ delete_elements_by_predicate(
     void* etc,
     struct r_set_cfg* cfg
 ) {
-    avl_dbg("Removing elements with predicate", NULL);
-
     // check whether the subtree is empty
     if (!*root) {
         return 0;
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -575,7 +575,7 @@ delete_elements_by_predicate(
         return 0;
     }
 
-    unsigned int retval=0;
+    unsigned int retval = 0;
 
     // iterate into subnodes
     retval += delete_elements_by_predicate(&(*root)->l, pred, etc, cfg);
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -559,3 +559,41 @@ find_node(
     return iter;
 }
 
+static unsigned int
+delete_elements_by_predicate(
+    struct avl_el** root,
+    r_predf pred,
+    void* etc,
+    struct r_set_cfg* cfg
+) {
+    avl_dbg("Removing elements with predicate", NULL);
+
+    // check whether the subtree is empty
+    if (!*root) {
+        return 0;
+    }
+
+    unsigned int retval=0;
+
+    // iterate into subnodes
+    retval += delete_elements_by_predicate(&(*root)->l, pred, etc, cfg);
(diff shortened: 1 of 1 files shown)


src/libreset/avl.c
```diff
@@ -255,7 +255,9 @@ avl_ndel(
     void* etc,
     struct r_set_cfg* cfg
 ) {
-    return delete_elements_by_predicate(&el->root, pred, etc, cfg);
+    unsigned int retval= delete_elements_by_predicate(&el->root, pred, etc, cfg);
```
matthiasbeyer: This shouldn't have passed the pre-commit hook. Missing space around `=` here.

src/libreset/avl.c
```diff
@@ -559,3 +561,41 @@ find_node(
...
+    struct r_set_cfg* cfg
+) {
+    avl_dbg("Removing elements with predicate", NULL);
+
+    // check whether the subtree is empty
+    if (!*root) {
+        return 0;
+    }
+
+    unsigned int retval=0;
```
matthiasbeyer: Same here.

src/libreset/avl.c
```diff
@@ -559,3 +561,41 @@ find_node(
...
 }
 
+static unsigned int
+delete_elements_by_predicate(
+    struct avl_el** root,
+    r_predf pred,
+    void* etc,
+    struct r_set_cfg* cfg
+) {
+    avl_dbg("Removing elements with predicate", NULL);
```
matthiasbeyer: Do we actually compile wil `DEBUG` set? So does this compile?

matthiasbeyer:
If you're sure about the upper comment, ACK for rebasing. Please ensure that it compiles!

src/libreset/avl.c
neithernut: With `printf` and other va_arg functions, you can spam the stack as much as you want. The function will not care as it only takes information about the arguments from the formatting string.
That's why they cause so many security issues in some programs.
The only thing which could prevent compilation are compiler warnings.
Frankly, if there is no way to print a message without any argument, the macro is buggy.

matthiasbeyer: Should we 'fix' the macro?

neithernut:
Mh... so much for that... the debug header is broken:
- It does not include `stdio.h`, which causes warnings/errors because of the implicit declaration of fprintf. (this can be solved by including `stdio.h` in line 49 of `debug.h`)
- It tried to print `__LINE__` using `%s` instead of `%d`

Also: a lot of debugging messages cause warnings/errors because of
- `rs_hash` is not an `int`
- missing casts from `struct something*` to `void*`
- too many arguments (caused by this PR)

Glibc (and maybe other libcs as well) ship with a header providing formatting string macros for all the types in `stdint.h` and other headers. Maybe we should use that, if it's availible.

So... what shall we do about this? Fixing the debug-header is ovbiously nothing that would go in this branch. But what should I do about the paramerterless debugging message? Ommit the debugging message completely?

matthiasbeyer:
Remove the debug printing from this PR and open a new issue with all information (and maybe more?) you posted above. Feel free to implement an alternative and create fixup patches (in one big PR).

neithernut:
I'll do so once I'm done with my current stuff.

matthiasbeyer:
:+1:

neithernut:
Done.

Issue #159, opened 2014-09-19
Naming of insert functions not consistent.


matthiasbeyer:
We have:
- `ht_insert()`
- `avl_add()`
- `ll_insert()`

We should be consistent here.

TheNeikos:
Yes, because you insert into a list, but you don't insert into an AVL tree as you don't really know where it will end up. But I guess renaming it shouldn't be a problem.

neithernut:
We,, we changed the semantics a little a few days ago.
Before, it was _adding_ a node.
Now it's _inserting_ an element.
(the AVL and the linked list are, in a way, each a set implementation of their own)

neithernut:
This may now be closed.

PR #162, opened 2014-09-19
Renamed function: avl_add() -> avl_insert()

matthiasbeyer:
Solves #159


neithernut:
ACK.

neithernut:
Mh... also: `avl_insert()` should not return the node, but either `1` or `0`.

neithernut:
I suggest merging this PR, I will then rebase #163.

PR #177, opened 2014-09-20
Add attributes to set header

matthiasbeyer:


matthiasbeyer:
I don't know why this fails. @neithernut can you tell me why? I guess this has something to do with the build env.

neithernut:
The include paths do not include the util directory iself.
Is `attributes.h` a public header?
If not, it should also not be included in `set.h`.

matthiasbeyer:
@neithernut So I'll close this PR unmerged.

PR #184, opened 2014-09-21
Added implementation for r_set_remove()

matthiasbeyer:
Targets #167 .

PR #181, opened 2014-09-21
Added implementation for r_set_contains()

matthiasbeyer:
Targets #168

PR #185, opened 2014-09-21
Added (missing) entry for set.c in CMakeLists.txt

matthiasbeyer:
Should be merged before #181, #182, #183 and #184, so they can be
rebased onto this and so the CI actually test compiles patches
in these PRs, too.


neithernut:
ACK.

PR #188, opened 2014-09-22
Add cleared pages between sections

neithernut:
This should make tha paper a bit prettier.


matthiasbeyer:
ACK. No cleanup required as this is only one commit.

Issue #186, opened 2014-09-22
Bug: regen_metadata() and rotation causing node count to decrease where it should increase


TheNeikos:
As outlined in #156, it seems to me that `avl_node_cnt` and `avl_insert` do not play well together.

The relevant code is:

``` c
for (i = 0; i < 10; i++) {
    ck_assert_int_eq(1, avl_insert(avl, hash[i], &data[i], &cfg_int));                          
}
ck_assert_int_eq(avl_node_cnt(avl->root), 10);
```

Unless I am not understand an avl tree, if I insert 10 different data values with 10 different hashes, I should get 10 nodes. However, `avl_node_cnt` returns 2 consistently. Is this expected behaviour?

matthiasbeyer:
I did some debugging on this.

It seems that `regen_metadata()` does not work properly. To reproduce (gdb):
- Start gdb and run until `test_avl_insert_multiple`
- Stop execution before first insert, should be line 54
- Watch `avl->root->node_cnt` variable

```
break test_avl_insert_multiple
run
break 54
c
watch avl->root->node_cnt
```

The hit `c` to continue several times. The node count increases like so:
`<unreadable> -> 1`, `1 -> 2`, `2 -> 3` and then jumps: `3 -> 1`. Always in `regen_metadata()`. Watch at the backtrace (`bt`) between the continues (`c`), to see the stack. It always _adds_  nodes, so the node count should _increase_. But it jumps from 3 to 1, which means that `avl_node_cnt()` on avl.c#507 results either in negative numbers or returning wrong stuff (I would assume the latter, as we're talking of `unsigned int` here everywhere...).

I guess this error is due to rotation, I think the tree gets rotated and the updating of the node count does not work properly (will try to find this out with the debugger now).

---

Update:

avl.c#336 causes the node count to decrease.

matthiasbeyer:
Closed by #190

PR #190, opened 2014-09-22
Fix AVL insert

neithernut:
The result of the `rebalance_subtree()` operation in `avl_insert()` was ignored. Which is bad, since the result is the new root of the AVL.
Seems to "solve" issue #186 (now it crashes, complaining about double-free'ing some int in `test_avl_insert_multiple()`).


matthiasbeyer:
Tested. Solved the issue from #186 . Merged.

PR #191, opened 2014-09-22
Fix: `_find()` functions shouldn't modify parameters, make them const

matthiasbeyer:
Because of this I'm having a hard time implementing `r_set_contains()`
(for example).


neithernut:
ACK.

neithernut:
But why did you attach the "bug" label to _this_?

matthiasbeyer:
"Bug" label was added by accident, sorry.

PR #192, opened 2014-09-22
Fix: Destroying datastructures with const cfg

matthiasbeyer:


neithernut:
ACK. `cfg` should, in fact, always be const.

PR #194, opened 2014-09-22
Fix: The `_del()` and `_ndel()` operations shouldn't be able to alter cfg or cmp element

matthiasbeyer:
The name of this branch maybe is confusing, but follows the naming schemes
of the other PRs which do similar work.

### DIFF
--- a/src/libreset/avl.c
+++ b/src/libreset/avl.c
@@ -239,7 +239,7 @@ avl_ndel(
     struct avl* el,
     r_predf pred,
     void* etc,
-    struct r_set_cfg* cfg
+    struct r_set_cfg const* cfg
 ) {
     unsigned int retval = delete_elements_by_predicate(&el->root, pred, etc, cfg);
     el->root = rebalance_subtree(el->root);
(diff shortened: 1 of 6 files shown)


src/libreset/avl.c
```diff
@@ -163,8 +163,8 @@ static unsigned int
 delete_elements_by_predicate(
     struct avl_el** root,
     r_predf pred,
-    void* etc,
-    struct r_set_cfg* cfg
+    void const* etc,
```
matthiasbeyer: This causes the compilation error.

When we define the `etc` parameter for the predicate function as `const`, we must also alter the predicate function type definition.

Is this feasible? Or should the `etc` parameter for the predicate function type be alterable?

neithernut: As stated some time ago: the predicate should be able to alter the `etc` parameter.

matthiasbeyer: Ok. Then I will submit fixups for this.

neithernut:
ACK.

matthiasbeyer:
Squashing done, waiting on CI.

neithernut:
This may now be merged.

PR #196, opened 2014-09-22
Fix: The hash function should not alter the data

matthiasbeyer:


matthiasbeyer:
Helps #181 , #182 , #183 , #184

TheNeikos:
I think I can ack this

matthiasbeyer:
I would like to get an ack from @neithernut , too.

neithernut:
ACK.

Issue #199, opened 2014-09-23
Speed up travis.


matthiasbeyer:
Travis gets painful slow. We should look into the speedup options.

neithernut:
Travis CI docs on caching: http://docs.travis-ci.com/user/caching/
It sais: "only available for private repositories on travis-ci.com"

matthiasbeyer:
Damn.

matthiasbeyer:
Maybe this can help: http://docs.travis-ci.com/user/speeding-up-the-build/

---

The actual problem is not the build (well, the latex stuff takes some time...) but the installing of the dependencies. I guess we have no luck here.

TheNeikos:
Running our own solution?

TheNeikos:
In the end all you need is a hook that sends a ping to a server and which then git clones and runs a shell script and sends back the return value of the script.

neithernut:
Indeed.

neithernut:
May now be closed due to #204.

matthiasbeyer:
More likely due to #200 , as we are still on travis. But yes.

PR #200, opened 2014-09-23
Add build env speedup

neithernut:
This PR will hopefully help solving issue #199.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,6 @@ install:
 
 script:
     - cmake -DHARD_MODE=ON .
-    - make
-    - make paper
+    - make -j3
+    - make -j3 paper
     - ctest -VV .


matthiasbeyer:
As the build does build the paper _while_ building the sources. Wouldn't it be better to build the source with `-j3` and _afterwards_ the paper with `-j3`?

neithernut:
If we want to look at the Travis CI output on a regular basis, this would indeed be the better option.
However, we use local checks, either through hooks or by manually trying to compile. For this case, the additional speedup of doing everything in _one_ make call is preferable imho.

neithernut:
I split up the make calls.

matthiasbeyer:
Feel free to rebase now.

neithernut:
Done.

PR #202, opened 2014-09-23
Reorganize the source according to #197

### LINKED ISSUE #197: How should the set operations be implemeneted on the lower levels?
How should the architecture look like for the lower parts of the library, namingly the hashtable, avl and linked list implementation?

For example, if you do an union between two sets, does the set code handle all this, or are there sub-problems which can be solved on the level of the hash table or avl?

---

Also, but not that important: We _could_ split up the code for the set function implementations in own files under `src/libreset/operations/*.c` to keep things less complex. I assume there will be a lot of code for these functions, so it would be nice to have them split up.

For common code parts, we could create a `src/libreset/operations/common.{c,h}`, where common code snippets would resist then.

What do you think?

matthiasbeyer:
This closes #197 .

This PR reorganizes the source code accordingly to #197.
@neithernut , @manuelmessner , @TheNeikos I would like to get this
reviewed _really_ carefully.


neithernut:
The history was not preserved.

matthiasbeyer:
I'll redo the whole history, as this is really important for me.

matthiasbeyer:
I think I am not able to preserve the history. Feel free to try it on your own, @neithernut , but I am fine with the current history of the PR if there is no other issue besides the history (though it would be nice to preserve the history).

neithernut:
I'll give it a try.

PR #205, opened 2014-09-24
The conditional indent setting for astyle was wrong

matthiasbeyer:
This:

``` C
if (foo ||
    bar) {
    baz();
}
```

is not really readable. Format like so:

``` C
if (foo ||
        bar) {
    baz();
}
```


neithernut:
ACK.

PR #208, opened 2014-09-24
Fix: Use ll_find() in avl_find()

matthiasbeyer:
Solves todo from source.


neithernut:
Ah, yes. I totally forgot about that TODO.

neithernut:
(Implies ACK).

Issue #101, opened 2014-09-16
Testing: HT 


matthiasbeyer:
Implement Hashtable testing

Should be done _before_ the actual hashtable implementation, and should be merged onto an extra branch for the hashtable implementation, so we can slowly move to a n-branch-workflow.

matthiasbeyer:
We should/could do a test-driven approach here: One who implements a functionality can first write the test and then implement the functionality required to pass the test.

Relates #102 , #62 , #63

neithernut:
May now be closed.

PR #210, opened 2014-09-24
Implement ht tests

matthiasbeyer:
Targets #101

### DIFF
--- a/tests/ht/ht_tests.c
+++ b/tests/ht/ht_tests.c
@@ -16,6 +16,21 @@ START_TEST (test_ht_init) {
 }
 END_TEST
 
+START_TEST (test_ht_init_big) {
+    size_t exp;
+
+    for (exp = 2; exp < 10; exp++) {
+        struct ht ht;
+        ht_init(&ht, exp); /* allocate 2^exp */
+
+        ck_assert(ht.buckets != NULL);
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
@neithernut We have a _really_ interesting problem here.

Try to run the tests. The tests fail for the following values:
- Number of inserted values: 1000
- Number of buckets in the hashtable: 2^2 (4) up to 2^8 (256).

So the idea comes up that it works while the number of buckets is greater than the half of the number of inserted values (it does _not_ fail for 512 buckets and 1024 buckets).

**But** it also does not fail when having 4 buckets and inserting 100 values. Obviously 4 < (100/2). So this seems to be a _really_ strange bug. Maybe _you_ can trace it, but I am unable to find it and neither does @TheNeikos  .

---

Another good point: I insert the _last_ value as first, so it fails when inserting `1000`, which is the `first` value inserted!

neithernut:
I'll look into this as soon as I can spare the time.

matthiasbeyer:
nah, @TheNeikos and I fixed this, I guess (can't remember, ... I'm still a bit tired, sorry...)

matthiasbeyer:
@neithernut , @TheNeikos SYN

TheNeikos:
ACK-SYN

neithernut:
ACK.

PR #211, opened 2014-09-24
Add r_set_cardinality()

TheNeikos:
This will add `r_set_cardinality` functionality with tests for each layer.


src/libreset/avl/avl.h
```diff
@@ -131,6 +131,20 @@ avl_find(
 );
 
 /**
+ * Get the cardinality of this avl tree (number of elements, not nodes)
+ *
+ * @memberof avl
+ *
+ * @return The cardinality
```
matthiasbeyer, line 140: Be a bit more verbose here, please.

src/libreset/avl/avl_cardinality.c
```diff
@@ -0,0 +1,25 @@
...
+
+unsigned int
+avl_cardinality(
+    struct avl const* avl
+) {
+    if(!avl) {
+        return 0;
+    }
+
+    return subtree_cardinality(avl->root);
```
matthiasbeyer, line 24: You can implement the whole function with a ternary operator (`?:`), which results in 1 line instead of 5.

src/libreset/ll/ll_count.c
```diff
@@ -0,0 +1,21 @@
...
+) {
+    if(!(ll || ll->head)) {
+        return 0;
+    }
+
+    int size = 0;
+    struct ll_element* iter = ll->head;
+
+    while(iter)
+    {
```
matthiasbeyer: This was obviously not run through astyle. The brace has to be on the same line as the `while` statement!

matthiasbeyer:
Nice and good structured PR! I like your patches!

src/libreset/avl/avl_cardinality.c
neithernut, line 24: We already discussed that this is not advisable and might even result in crashes.
Keep the `if`.

TheNeikos:
This can now be reviewed.

neithernut:
ACK.

PR #215, opened 2014-09-24
Added avl_insert_many test

TheNeikos:
This adds a test with 10000 inserted elements.

### DIFF
--- a/tests/avl/avl_tests.c
+++ b/tests/avl/avl_tests.c
@@ -62,6 +62,25 @@ START_TEST (test_avl_insert_multiple) {
 }
 END_TEST
 
+START_TEST (test_avl_insert_many_distinct) {
+    struct avl* avl = calloc(1, sizeof(*avl));
+
+    int data[10000];
+
+    int i;
+    for (i = 0; i < 10000; i++) {
+        data[i] = i;
+        ck_assert(1 == avl_insert(avl, data[i], &data[i], &cfg_int));
+    }
+
+    for (i = 0; i < 10000; i++) {
+        ck_assert(&data[i] == avl_find(avl, data[i], &data[i], &cfg_int));
+    }
+
+    ck_assert(1 == avl_destroy(avl, &cfg_int));
+}
+END_TEST
+
 START_TEST (test_avl_insert_multiple_destroy) {
     struct avl* avl = calloc(1, sizeof(*avl));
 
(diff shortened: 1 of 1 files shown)


tests/avl/avl_tests.c
```diff
@@ -62,6 +62,25 @@ START_TEST (test_avl_insert_multiple) {
 }
 END_TEST
 
+START_TEST (test_avl_insert_many) {
```
matthiasbeyer: I'd like you to rename this to `test_avl_insert_many_distinct`, as there does no collision happen with this test and there should be tests for collision checking. So the naming shows that this test is a simple one.

matthiasbeyer:
Please squash your commits, I'm ready for merge afterwards.

tests/avl/avl_tests.c
```diff
@@ -62,6 +62,25 @@ START_TEST (test_avl_insert_multiple) {
 }
 END_TEST
 
+START_TEST (test_avl_insert_many_distinct) {
+    struct avl* avl = calloc(1, sizeof(*avl));
+
+    int data[10000];
```
neithernut: If you want a "huge" bunch of ints, you should allocate it on the heap.
Also, you should probably fill the integers using some random number generator.

tests/avl/avl_tests.c
```diff
@@ -62,6 +62,25 @@ START_TEST (test_avl_insert_multiple) {
...
+
+    int i;
+    for (i = 0; i < 10000; i++) {
```
neithernut: Maybe parameterize the number of ints (e.g. use a define).
Also: would count down rather than up.

tests/avl/avl_tests.c
matthiasbeyer: Nope, not random. We had a kind-of hard time this night with randomness in tests, so it'll be better using non-randoms!

neithernut: The solution used in the loop is acceptable.

neithernut:
ACK.

matthiasbeyer:
@TheNeikos wanna squash something?

TheNeikos:
I can if you want me to

matthiasbeyer:
You're the author. You _can_ rebase now, if you want. I don't think it is neccessary, but it is your decision!

TheNeikos:
Then I don't want to.

PR #216, opened 2014-09-24
Added simplification for bucket index equation

matthiasbeyer:
The problem with the "old" equation was, that you cannot implement it in
C as-is. You have to transform the equation into the form which is added
by this PR to get it compiling.

For referencing, this simplification should be added in the paper, too.

### DIFF
--- a/paper/data-layout.tex
+++ b/paper/data-layout.tex
@@ -26,6 +26,7 @@
 
         \begin{equation}
             s \in \left\{ 2^n | n \in \mathbb{N} \right\}
+            \label{eq:num_of_buckets}
         \end{equation}
 
         Assume having two hashtables $A$ and $B$, where the number of buckets of
@@ -64,6 +65,12 @@
 
         \begin{equation}
             i \in e(b_n) \qquad j = h(i) \div \frac{2^N}{s}
+        \end{equation}
+
+        Using equation \ref{eq:num_of_buckets}, this can be transformed to:
+
+        \begin{equation}
+            j = h(i) \div \frac{2^N}{2^n} = h(i) \div 2^{N - n}
         \end{equation}
 
         Where $N$ is the number of bits of a hash-value.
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
@neithernut

paper/data-layout.tex
```diff
@@ -24,7 +24,7 @@ \section{Data layout}
         So far, we're just talking about an average hashtable. However, we
         constrain the number of buckets $s$ to powers of two:
 
-        \begin{equation}
+        \begin{equation} \label{eq:num_of_buckets}
```
neithernut: In the other equation I labeled, I put the label in a seperate line just before the `\end{equation}`. Please do the same here.

paper/data-layout.tex
```diff
@@ -63,7 +63,13 @@ \section{Data layout}
         However, we are using the \em most\em{} significant bits:
 
         \begin{equation}
-            i \in e(b_n) \qquad n = h(i) \div \frac{2^N}{s}
+            i \in e(b_n) \qquad n_i = h(i) \div \frac{2^N}{s}
```
neithernut: The `n_i =` is wrong. It's plain `n =`.

paper/data-layout.tex
```diff
@@ -63,7 +63,13 @@ \section{Data layout}
...
+        \end{equation}
+
+        Using equation \ref{eq:num_of_buckets}, this can be simplified to:
```
neithernut: It's not a simplification, but a transformation.

paper/data-layout.tex
matthiasbeyer: The `n` variable then conflicts with the `2^n` from the equation from the size of the hastable.

paper/data-layout.tex
matthiasbeyer: Good point, will patch this.

paper/data-layout.tex
neithernut: Then please use a different character, but do not add an index (which, on top, is used in the same equation).

matthiasbeyer:
@neithernut .

neithernut:
Acceptable.

PR #217, opened 2014-09-24
Added a missing bitcount operator

TheNeikos:
This fixes a bug in `ht_insert`.

This bug made it really unclear which bucket was selected. If it worked before than it was pure luck, as it was undefined behaviour.

### DIFF
--- a/src/libreset/ht/base.c
+++ b/src/libreset/ht/base.c
@@ -76,6 +76,6 @@ ht_insert(
     struct r_set_cfg const* cfg
 ) {
     rs_hash hash = cfg->hashf(data);
-    size_t i = hash >> (sizeof(hash) - ht->sizeexp);
+    size_t i = hash >> (BITCOUNT(hash) - ht->sizeexp);
     return avl_insert(&ht->buckets[i].avl, hash, data, cfg);
 }
--- a/src/libreset/ht/base.c
+++ b/src/libreset/ht/base.c
@@ -76,6 +76,9 @@ ht_insert(
     struct r_set_cfg const* cfg
 ) {
     rs_hash hash = cfg->hashf(data);
+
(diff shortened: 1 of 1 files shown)


src/libreset/ht/base.c
```diff
@@ -76,6 +76,6 @@ ht_insert(
     struct r_set_cfg const* cfg
 ) {
     rs_hash hash = cfg->hashf(data);
-    size_t i = hash >> (sizeof(hash) - ht->sizeexp);
+    size_t i = hash >> (BITCOUNT(sizeof(hash)) - ht->sizeexp);
```
matthiasbeyer: `BITCOUNT()` gets a type and uses it with `sizeof()` internally.

matthiasbeyer: There are at least 10 lines documentation missing here.

src/libreset/ht/base.c
```diff
@@ -76,6 +76,6 @@ ht_insert(
     struct r_set_cfg const* cfg
 ) {
     rs_hash hash = cfg->hashf(data);
-    size_t i = hash >> (sizeof(hash) - ht->sizeexp);
+    size_t i = hash >> (BITCOUNT(hash) - ht->sizeexp);
```
matthiasbeyer, line 82: Please add documentation to this line.

matthiasbeyer:
@neithernut Are these patches correct?

neithernut:
ACK. (The `sizeof(hash)` was malicious)

neithernut:
Maybe we should put the calculation of the bucket's address/id from a hash into a seperate function. (We will need this in a few places.)

matthiasbeyer:
@neithernut Indeed. We don't need an issue for this, so @TheNeikos, do you want to submit a PR for this (separate from this one, of course)?

Issue #176, opened 2014-09-20
Implement: r_set_select()


matthiasbeyer:

TheNeikos:
This can now be closed

PR #223, opened 2014-09-25
Implementation of r_set_select()

matthiasbeyer:
Targets #176

### DIFF
--- a/src/libreset/ht/ht.h
+++ b/src/libreset/ht/ht.h
@@ -174,6 +174,24 @@ __r_nonnull__(1, 3, 4)
 ;
 
 /**
+ * Select entries from a ht into a new one
+ *
+ * @memberof ht
+ *
+ * @return zero on success, else error code
+ */
+int
+ht_select(
+    struct ht* src, //!< Destination hashtable object
+    r_predf pred, //!< The predicate
+    void* pred_etc, //!< Additional information for the predicate function
+    r_procf procf, //!< function processing the selected values
+    void* dest //!< some pointer to pass to the procf
+)
+__r_nonnull__(1, 4)
+;
+
+/**
  * Helper to calculate the actual bucket count of the hashtable
  *
  * @memberof ht
--- a/src/libreset/ll/ll.h
+++ b/src/libreset/ll/ll.h
@@ -191,6 +191,24 @@ __r_warn_unused_result__
     for (struct ll_element* it = (ll)->head; it; it = it->next)
 
 /**
+ * Select entries from a linked list into a new one
+ *
+ * @memberof ll
+ *
+ * @return zero on success, else error code
+ */
+int
+ll_select(
+    struct ll const* src, //!< The source from where to select
+    r_predf pred, //!< The predicate
+    void* pred_etc, //!< Additional information for the predicate function
+    r_procf procf, //!< function processing the selected values
+    void* dest //!< some pointer to pass to the procf
+)
+__r_nonnull__(1, 4)
+;
+
+/**
  * @}
  */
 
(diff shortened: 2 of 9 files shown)


src/libreset/ll/ll.h
```diff
@@ -167,6 +167,22 @@ __nonnull__(1)
...
+ * @memberof ll
+ *
+ * @return zero on success, else error code
+ */
+int
+ll_select(
+    struct ll const* src, //!< The source from where to select
+    r_predf pred, //!< The predicate
+    void* etc, //!< Additional information for the predicate function
+    int (*insf)(void* elem, void* insetc), //!< Function to insert somewhere
```
TheNeikos: `insetc` ?

matthiasbeyer:
The naming is not very good by now. I'm to tired to fix this now, will do tomorrow!

src/libreset/ht/ht.h
```diff
@@ -148,7 +164,7 @@ ht_ndel(
  */
 static inline size_t
 ht_nbuckets(
-    struct ht* ht //!< The hashtable object to calculate the bucket count for
+    struct ht const* const ht //!< The ht object to calc the bucket count for
```
TheNeikos, line 203: What is the connection to `r_set_select`?

src/libreset/ht/ht.h
```diff
@@ -161,7 +177,7 @@ ht_nbuckets(
 
 static inline size_t
 ht_nbuckets(
-    struct ht* ht
+    struct ht const* const ht
```
TheNeikos, line 218: What is the connection to r_set_select?

matthiasbeyer, line 218: Nothing. Just found this bug. Will extract tomorrow and submit in a separate PR.

neithernut:
The select function is supposed to put the selected elements into another set.

matthiasbeyer:
@neithernut yes? It does...?

neithernut:
I thought so...
Well, if we keep the vurrent definition, we should ensure that we can simply pass `set_insert()` as `insf`. This is currently not the case.

matthiasbeyer:
This would require the linked list implementation to include the `libreset/set.h` header, which I consider as very ugly.

There for the helper function, which has a more generic interface.

neithernut:
Nope, the function declaration just has to fit "coincidentally". (Which means that the function can be casted to the `insf` type)

matthiasbeyer:
@neithernut I do not understand your comment.

neithernut:
The parameters and return value has to be the same for `insf()` and `set_insert()` (except that `struct r_set*` is `void*` for `insf`).

matthiasbeyer:
But why, we pass a function which does insertion. I do the insertion into the ht here, as the set is absolutely not relevant... so it is much simpler to insert into the ht of the destination directly.

neithernut:
Yes, but if you want to give users the possibility to use a custom `insf()` to do _not_ use a set (like printing the selected elements directly), you should take care of the usecase of selecting elements into another set, too.
Also: the number of selected elements may potentially be quite low. In such cases, inserting one element after another does, indeed, make more sense.

matthiasbeyer:
I guess we shouldn't provide such an possibility.

neithernut:
Why not? This way we would need to implement the selection only once.
And the benefits of doing it twice are minimal in this case.

matthiasbeyer:
@neithernut can you send some commits for this?

neithernut:
Yep (after the switch to drone.io)

matthiasbeyer:
I'm ready for merge. @neithernut , @TheNeikos SYN.

neithernut:
Please rebase.

neithernut:
And squash.

neithernut:
ACK.

PR #227, opened 2014-09-25
Fix ll uninitialized values in test

TheNeikos:
This fixes a few memory leaks in the linked list tests as well as using calloc instead of malloc.


neithernut:
ACK.

PR #228, opened 2014-09-25
Rename prefixes

manuelmessner:
renamed:
rs_hash -> r_hash
RS_FORCE_INLINE -> R_FORCE_INLINE

PR #229, opened 2014-09-25
Fix: ht bucket index calculation

matthiasbeyer:


matthiasbeyer:
Waiting on #228

src/libreset/ht/base.c
```diff
@@ -3,6 +3,19 @@
 #include "ht/ht.h"
 #include "util/macros.h"
 
+/**
+ * Calculate the bucket index for a hash and a hashtable
+ *
+ * @return The index of the hash `hash` in the hashtable `ht`
+ */
+static inline size_t
```
TheNeikos, line 13: I would suggest using `R_FORCE_INLINE` here.

matthiasbeyer, line 13: Hm, would like to hear @neithernut s opinion.

neithernut, line 13: We might do this, _after_ the attribute header is fixed.

neithernut:
`bucket_index` could be forward declarated, as we did with some AVL functions.
Having all the internals at the end of the file is cleaner imho.

matthiasbeyer:
I don't want to add extra noise to the file with this, @neithernut .

neithernut:
k, then.
ACK.

PR #230, opened 2014-09-25
Add callgraph to doc of public interface

matthiasbeyer:


neithernut:
Not quite sure about this.

matthiasbeyer:
@manuelmessner , @TheNeikos

manuelmessner:
I don't think that we need this option.
If there is a function which we want to analyze, we can add the option temporary on the local machine.

PR #251, opened 2014-09-26
Add attributes to avl functions

matthiasbeyer:
This PR redoes #140

### DIFF
--- a/src/libreset/avl/avl.h
+++ b/src/libreset/avl/avl.h
@@ -64,6 +64,7 @@ avl_destroy(
     struct avl* avl, //!< The avl tree
     struct r_set_cfg const* cfg //!< type information proveded by the user
 )
+__r_nonnull__(1, 2)
 ;
 
 /**
@@ -82,6 +83,7 @@ avl_insert(
     void* const d, //!< The data element
     struct r_set_cfg const* cfg //!< type information provided by the user
 )
+__r_nonnull__(1, 3, 4)
 ;
 
 /**
@@ -102,6 +104,7 @@ avl_del(
     void const* cmp, //!< element to compare against
     struct r_set_cfg const* cfg //!< type information provided by the user
 )
+__r_nonnull__(1, 3, 4)
 ;
 
 /**
@@ -121,6 +124,7 @@ avl_ndel(
     void* etc, //!< User-data to pass to the predicate
     struct r_set_cfg const* cfg //!< type information provided by the user
 )
+__r_nonnull__(1, 3, 4)
 ;
 
 /**
@@ -137,6 +141,7 @@ avl_find(
     void const* const d, //!< The data element to compare to
     struct r_set_cfg const* cfg //!< type information proveded by the user
 )
+__r_nonnull__(1, 3, 4)
 __r_warn_unused_result__
 ;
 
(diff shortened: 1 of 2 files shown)


src/libreset/avl/avl.h
```diff
@@ -197,6 +204,7 @@ inline static unsigned int
 avl_node_cnt(
     struct avl_el const* root //!< The root element
 )
+__r_nonnull__(1)
```
neithernut: This function includes a `NULL`-check.

src/libreset/avl/avl.h
```diff
@@ -182,6 +188,7 @@ inline static unsigned int
 avl_height(
     struct avl_el const* root //!< The root element
 )
+__r_nonnull__(1)
```
neithernut: This function includes a `NULL`-check.

src/libreset/avl/avl.h
matthiasbeyer: This macro is not a runtime check. So we need both, runtime check for attributes being `NULL` and compiletime checks.

neithernut: But if the function has a `NULL` check _by design_ (because it is used in recursive algos) it's bogous to add the `nonnull` attribute.
The thing is: attributes are also a form of annotation/documentation. If a function has a `NULL`-check, it shouldn't be attributed with `nonnull`.

matthiasbeyer: With your proposal, there are two cases:

1) The code does a `NULL`-check on its own. Using a function from the outside (from outside the library) can lead to a runtime error, as if someone passes `NULL` accuidentially, the code recognizes this and returnes with failure.

2) The code has no internal `NULL` checks, so we cannot accidentially compile something which passes `NULL` somewhere, but the user can. This would result in a segfaults somewhere.

Having both the runtime-`NULL`-checks and the compiletime checks prevents as much errors as possible. So why not having it?

neithernut: Because you say that a function will not accept a `NULL`-pointer when, in fact, you depend on it accepting `NULL`-pointers --at least for those two functions.
Worst thing that could happen would be someone saying "Ah, that function should not take `NULL`-pointers. Let's optimise it by removing the check!".

matthiasbeyer: We already had a case where the `__nonnull__` attribute was missing and the function did crash because of it. So I would highly apprechiate keeping both the attribute and the check. It is also not that much code.

neithernut: Yes, but in _this case_ we _depend_ on the function accepting `NULL`. Read `regen_metadata()`.

src/libreset/avl/common.h
```diff
@@ -15,7 +15,9 @@ void
 destroy_subtree(
     struct avl_el* node, //!< A node to destroy
     struct r_set_cfg const* cfg //!< type information proveded by the user
-);
+)
+__r_nonnull__(1, 2)
```
neithernut: `node` may be `NULL`. It's a completely sane value to pass to this function.

src/libreset/avl/common.h
```diff
@@ -30,6 +32,7 @@ struct avl_el*
 rebalance_subtree(
     struct avl_el* root //!< The root of the subtree to rebalance
 )
+__r_nonnull__(1)
```
neithernut: Same here. The `nonnull` should be omitted.

neithernut:
ACK.

PR #260, opened 2014-09-26
Modify avl_find to return a pointer to a pointer

neithernut:
This PR modifies `avl_find()`. While previously it returned a pointer to a node or NULL, now it will return a two-level pointer.
This is a prequisite to my work on the `avl_union()` algorithm.

### DIFF
--- a/src/libreset/avl/common.c
+++ b/src/libreset/avl/common.c
@@ -213,26 +213,30 @@ regen_metadata(
     }
 }
 
-struct avl_el*
+struct avl_el**
 find_node(
     struct avl const* avl,
     r_hash hash
 ) {
     avl_dbg("Finding node with hash: 0x%zx", hash);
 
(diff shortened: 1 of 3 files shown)


src/libreset/avl/common.c
```diff
@@ -213,26 +213,30 @@ regen_metadata(
...
-struct avl_el*
+struct avl_el**
 find_node(
     struct avl const* avl,
     r_hash hash
 ) {
     avl_dbg("Finding node with hash: 0x%zx", hash);
 
-    struct avl_el* iter = avl->root;
+    struct avl_el** iter = (struct avl_el**) &avl->root;
```
matthiasbeyer, line 223: Do you really need this cast?

neithernut:
I found another, cleaner solution to the problem.

matthiasbeyer:
:+1:

PR #263, opened 2014-09-27
Redefine r_set_select()

neithernut:
This PR brings with it a complete redefinition of `r_set_select()`.
It is now possible to process selected elements using a user defined function or insert selected elements into another set by passing `r_set_insert` as `procf` and the destination set as `dest`.

### DIFF
--- a/include/libreset/set.h
+++ b/include/libreset/set.h
@@ -55,6 +55,20 @@ typedef int (*r_predf)(void const*, void*);
 
 
 /**
+ * Insertion/processing function
+ *
+ * Use functions of this type if you want to select or process elements of a
+ * set.
+ * The function is expected to return 0 on success and negative values on error.
+ * The first parameter is considered a user defined value (e.g. another set to
+ * insert those values), while the elements will be fed to the function as the
+ * second parameter.
+ */
+typedef int (*r_procf)(void*, void const*);
+
+
+
+/**
  * Allocate and initialize set object
  *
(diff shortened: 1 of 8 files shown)


include/libreset/set.h
```diff
@@ -55,6 +55,20 @@ typedef int (*r_predf)(void const*, void*);
...
+ * Insertion/processing function
+ *
+ * Use functions of this type if you want to select or process elements of a
+ * set.
+ * The function is expected to return 0 on success and negative values on error.
+ * The first parameter is considered a user defined value (e.g. another set to
+ * insert those values), while the elements will be fed to the function as the
+ * second parameter.
+ */
+typedef int (*r_procf)(void*, void const*);
```
matthiasbeyer, line 67: Wouldn't it make sense to switch the parameter order, as all other functions a user can pass to our library (e.g. through the config) gets the extra parameter as last one?

neithernut, line 67: Nope, because the way it is, you can pass `r_set_insert()` and similar functions, where the context comes _first_.
Have a look at libraries like https://github.com/clibs/inih. They also have a way to pass user defined stuff as the first parameter. I once used that library to parse the configuration and was able to pass the method to ad values to the config as-is, without a wrapper to change the order of arguments.

matthiasbeyer, line 67: Sounds resonable to me.

Issue #246, opened 2014-09-26
Remove astyle


matthiasbeyer:
As @neithernut stated in #243 , astyle removes labels from the code.

Remove it. NOW.

matthiasbeyer:
@neithernut Can this one be closed due to #264 ? If yes, you're welcome to close this issue and mark the milestone (which is due tomorrow) as ready afterwards.

Issue #238, opened 2014-09-25
Make compile and test step optional in pre-commit hook?


matthiasbeyer:
We should make the compile and testing step optional in the pre-commit hook, as we have a fast CI now.

What do you think?

neithernut:
We _may_ support disabling some stuff using environment variables (or some other technique).
However, using the commit hooks for compiling locally has two advantages:
- _Every_ commit is checked.
- You save yourself one roundtrip.

matthiasbeyer:
Of course it has, but I could understand if contributors with slower machines would prefer the additional roundtrip!

TheNeikos:
Without the debug output the compiling/testing should not be a problem I believe.

PR #264, opened 2014-09-27
pre-commit: Make astyle step optional

matthiasbeyer:


matthiasbeyer:
Relates #238 , but doesn't target it.

matthiasbeyer:
Targets #246 .

hooks/pre-commit
```diff
@@ -44,6 +34,30 @@ die() {
...
 
+astyle=0 # default: astyle OFF
+if [[ $PRE_COMMIT_ASTYLE =~ ^(y|Y)$ ]];
+then
+    astyle=1
+fi
+
+if ! type astyle 2> /dev/null
+then
+    if [[ $PRE_COMMIT_ASTYLE =~ ^(y|Y)$ ]] && [[ $astyle == 1 ]]; then
```
neithernut: You check the env here, too? You did that on line 38 already.

matthiasbeyer: Yes, but on lines 38 I check if it should be enabled. Here I check "If astyle is not available but should be used"

neithernut: Er... no. You check whether the env is set and whether the variable `astyle` is one. The former one can be omitted, since `astyle` is `0` if it's not set in line 40.
I think you confuse this with line 43.

neithernut:
ACK.

PR #266, opened 2014-09-28
Fix: CONSTPOW_TWO() macro type

matthiasbeyer:


TheNeikos:
ACK

neithernut:
When merging this, we have to change every "call" of this macro: we have to cast the operand to `size_t`.

matthiasbeyer:
Then it is better if we do not merge this, right?

TheNeikos:
I think it is only used for ht->sizeexp which is _already_ a `size_t`

matthiasbeyer:
So I'll close this now.

Issue #128, opened 2014-09-17
errno.h for error values


matthiasbeyer:
Use errno.h header constants for error return values.

neithernut:
Can this be closed?

matthiasbeyer:
I think so.

Please check the public interface whether there are errno returns missing.

matthiasbeyer:
@neithernut , @TheNeikos Is this done? Can you check the code for missing errno returns? (I'll do as well, of course)

matthiasbeyer:
I think this can be closed now.

PR #269, opened 2014-09-28
Update the return for returning errno consts

matthiasbeyer:
The indention of the PR was to check that every function returns errno (if
it should return a status).

Actually, only one line in the documentation was wrong, which is fixed by
this PR.


matthiasbeyer:
closes #128 .

neithernut:
ACK.

PR #271, opened 2014-09-28
Fix c/p error in cmake/module

matthiasbeyer:


TheNeikos:
ACK

TheNeikos:
What is interesting is that it build anyway.

neithernut:
ACK.

PR #279, opened 2014-10-09
Fix: Logical OR in ll_select()

matthiasbeyer:
If `pred` is `NULL`, the line still tries to execute the function. This
is fixed by this commit.


TheNeikos:
Please provide a test case that checks this behaviour.

neithernut:
ACK.

PR #280, opened 2014-10-09
We want to call the processing function if pred is NULL

neithernut:
This fixes the fix introduced my Matthias Beyer.
Which I was dump enough to merge.


matthiasbeyer:
ACK.

PR #282, opened 2014-10-13
Fix: The predicates parameter for the current element should not be const

matthiasbeyer:
The element which gets passed to the processing function shouldn't be `const`.

This PR fixes this.


neithernut:
ACK.

